PR #2, opened 2019-03-23
memory related updates

yatli:
1. add helpers `select_bank` and `memcpy_bank1`
2. move memory layout specifications all into MCLMemory.h; update memory layout.

### DIFF
--- a/avr/cores/megacommand/MCL/GridRowHeader.h
+++ b/avr/cores/megacommand/MCL/GridRowHeader.h
@@ -3,14 +3,14 @@
 #ifndef GRIDROW_H__
 #define GRIDROW_H__
 
-#define GRID_WIDTH 22
-
 class GridRowHeader {
+ private:
+  static constexpr int c_GridWidth = 22;
  public:
   bool active;
   char name[17];
-  uint8_t track_type[GRID_WIDTH];
-  uint8_t model[GRID_WIDTH];
+  uint8_t track_type[c_GridWidth];
+  uint8_t model[c_GridWidth];
 
   void update_model(int16_t column, uint8_t model_, uint8_t track_type_);
   bool write(int16_t row);
--- a/avr/cores/megacommand/MidiUart.cpp
+++ b/avr/cores/megacommand/MidiUart.cpp
@@ -277,14 +277,12 @@ bool MidiUartClass::avail() { return !rxRb.isEmpty(); }
 uint8_t MidiUartClass::m_getc() { return rxRb.get(); }
 
 ISR(USART1_RX_vect) {
-  uint8_t old_ram_bank = switch_ram_bank(0);
+  select_bank(0);
   isr_midi();
-  switch_ram_bank(old_ram_bank);
 }
 ISR(USART2_RX_vect) {
-  uint8_t old_ram_bank = switch_ram_bank(0);
+  select_bank(0);
   isr_midi();
-  switch_ram_bank(old_ram_bank);
 }
 
 inline void isr_midi() {
@@ -436,7 +434,7 @@ inline void isr_midi() {
 #ifdef TX_IRQ
 ISR(USART1_UDRE_vect) {
   // uint16_t count = 0;
-  uint8_t old_ram_bank = switch_ram_bank(0);
+  select_bank(0);
   //  isr_midi();
   //  while (!MidiUart.txRb.isEmpty()) {
   if (!MidiUart.txRb.isEmpty()) {
@@ -456,11 +454,10 @@ ISR(USART1_UDRE_vect) {
   if (MidiUart.txRb.isEmpty()) {
     CLEAR_BIT(UCSR1B, UDRIE1);
   }
-  switch_ram_bank(old_ram_bank);
 }
 
 ISR(USART2_UDRE_vect) {
-  uint8_t old_ram_bank = switch_ram_bank(0);
+  select_bank(0);
   //  isr_midi();
 
   uint8_t c;
@@ -475,7 +472,6 @@ ISR(USART2_UDRE_vect) {
   if (MidiUart2.txRb.isEmpty()) {
     CLEAR_BIT(UCSR2B, UDRIE1);
   }
-  switch_ram_bank(old_ram_bank);
 }
 
 #endif
--- a/avr/cores/megacommand/WProgram.h
+++ b/avr/cores/megacommand/WProgram.h
@@ -88,4 +88,25 @@ extern inline uint8_t switch_ram_bank(uint8_t x) {
   return x;
 }
 
+#ifdef __cplusplus
+
+class RamBankSelector {
+  private:
+  uint8_t m_oldbank;
+  public:
+  RamBankSelector(uint8_t bank) { m_oldbank = switch_ram_bank(bank); }
+  ~RamBankSelector() { switch_ram_bank(m_oldbank); }
+};
+
+#define select_bank(x) RamBankSelector __bank_selector(x)
+
+#endif// __cplusplus
+
+#define memcpy_bank1(dst, src, len) \
+  do { \
+  switch_ram_bank(1); \
+  memcpy(dst, src, len); \
+  switch_ram_bank(0); \
+  } while (false)
+
 #endif /* WProgram_h */
--- a/avr/cores/megacommand/WProgram.h
+++ b/avr/cores/megacommand/WProgram.h
@@ -72,19 +72,12 @@ extern uint32_t write_count_time;
 extern uint16_t minuteclock;
 extern uint8_t ram_bank;
 extern inline uint8_t switch_ram_bank(uint8_t x) {
-  //USE_LOCK();
-  //SET_LOCK();
   uint8_t old_bank = (uint8_t) (PORTL >> PL6) & 0x01;
 
   if (x != old_bank) {
-    //DISABLE timer 1 if switching banks.
- //   if (x == 0) { sbi(TIMSK0, TOIE0); }
-  //  else { cbi(TIMSK0, TOIE0); }
     PORTL ^= _BV(PL6);
-  //  CLEAR_LOCK();
     return old_bank;
   }
-//  CLEAR_LOCK();
   return x;
 }
 
@@ -102,11 +95,10 @@ class RamBankSelector {
 
 #endif// __cplusplus
 
-#define memcpy_bank1(dst, src, len) \
-  do { \
-  switch_ram_bank(1); \
-  memcpy(dst, src, len); \
-  switch_ram_bank(0); \
-  } while (false)
+extern inline void memcpy_bank1(uint8_t dst, volatile void *src, uint32_t len) {
+  switch_ram_bank(1);
+  memcpy(dst, src, len);
+  switch_ram_bank(0);
+}
 
 #endif /* WProgram_h */
--- a/avr/cores/megacommand/main.cpp
+++ b/avr/cores/megacommand/main.cpp
@@ -200,7 +200,7 @@ bool in_irq = false;
 
 ISR(TIMER1_COMPA_vect) {
 
-  uint8_t old_ram_bank = switch_ram_bank(0);
+  select_bank(0);
 
   clock++;
   if (MidiClock.state == 2) {
@@ -220,7 +220,6 @@ ISR(TIMER1_COMPA_vect) {
     }
   }
   }
-  switch_ram_bank(old_ram_bank);
 }
 
 // XXX CMP to have better time
@@ -257,7 +256,7 @@ uint16_t minuteclock = 0;
 
 ISR(TIMER2_COMPA_vect, ISR_NOBLOCK) {
 
- uint8_t old_ram_bank = switch_ram_bank(0);
+  select_bank(0);
 
   slowclock++;
   minuteclock++;
@@ -277,7 +276,6 @@ ISR(TIMER2_COMPA_vect, ISR_NOBLOCK) {
   gui_poll();
 #endif
 
-  switch_ram_bank(old_ram_bank);
 }
 
 uint8_t sysexBuf[5500];
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.vscode
+bin
--- a/Changelog
+++ b/Changelog
@@ -6,6 +6,8 @@ Features and Fixes:
 current kit when sequencer is running. This fixes some complications with
 parameter locks, and also improves chain page responsiveness.
 
+Track muteGroups and trigGroups are now properly removed.
+
 - MIDI SDS Receive is fixed and working.
 
 - Added Loudness Page, and loudness features
--- a/MCL.ino
+++ b/MCL.ino
@@ -1,0 +1,1 @@
+#include "MCL.h"
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ https://www.arduino.cc/download_handler.php?f=/arduino-1.8.5-windows.zip
 
 2) Extract zip file to desktop.
 
-3) Download MIDICtrl20_MegaCommand and extract to the "arduino-1.8.5\hardware\" folder.
+3) Download MIDICtrl20_MegaCommand and extract to the `arduino-1.8.5\hardware\` folder. 
+
+4) If you have an admin-installed Arduino IDE, extract MIDICtrl20_MegaCommand to the arduino user directory `%USERPROFILE%\Documents\Arduino\hardware\` (e.g. `C:\Users\your_name\Documents\Arduino\hardware\`)
 
 
 ### Configure the Core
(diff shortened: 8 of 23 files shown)


avr/cores/megacommand/MCL/GridRowHeader.h
```diff
@@ -3,14 +3,14 @@
 #ifndef GRIDROW_H__
 #define GRIDROW_H__
 
-#define GRID_WIDTH 22
-
 class GridRowHeader {
+ private:
+  static constexpr int c_GridWidth = 22;
```
yatli, line 8: avoid GRID_WIDTH redefinition

yatli:
question: all the track data goes to bank1 (and shared parts in the internal memory), what's in bank0?

jmamma:
Let me explain the memory.

The ATMEGA2560 has 8KB of SRAM and is capable of addressing 64KB of memory at a time.

With the external SRAM connected  you get 8KB (internal) + 56KB (external) (8KB is always wasted per bank)

avr-gcc is used to specify the memory layout upon linking. 
https://www.nongnu.org/avr-libc/user-manual/malloc.html

The internal SRAM 8KB has been reserved for the program stack. This means, memory (variables + objects) defined in function calls will be limited to 8KB. Good thing about the stack is that the memory is instantaneously reclaimed when you exit a function. *edited

The remaining 56KB is bank selectable and is used by Global variables and Objects.
You could theoretically use this space for calls to malloc() but this is a bad idea on a small microprocessor, due to RAM fragmentation. It is better to layout all the memory you are going to use ahead of time, and dont change it.


Bank Switching:

The MC features 2 banks. Bank 0 and Bank 1. Bank 0 is the default bank is used to store data that is essential to program function. 

Some examples of data stored in Bank 0: The MIDI and SYSEX buffers. The sequencer data. These data structures are accessed in ISR()s and should be as quick to access as possible.

Bank 1, is mainly used for cached data. For example, in chain mode, tracks are transferred from the SD card to bank 1, for faster load times.

Because many of the MIDICTRL data structures are stored Globally, switching banks and then trying to access these structures will cause the MC to crash. You need to be more clever. When switching banks, only memory in the stack* is reliable.
Data is thus syphoned in and out of second memory bank to an object in the stack*, like so:

```
379 bool MDTrackLight::load_from_mem(uint8_t column, uint32_t region) {
380   uint32_t len = sizeof(GridTrack) + sizeof(MDSeqTrackData) + sizeof(MDMachine);
381 
382   uint32_t pos = region + len * (uint32_t)column;
383 
384   volatile uint8_t *ptr;
385 
386   ptr = reinterpret_cast<uint8_t *>(pos);
387   // cbi(TIMSK0, TOIE0);
388   switch_ram_bank(1);
389   memcpy(this, ptr, len);
390   switch_ram_bank(0);
391   // sbi(TIMSK0, TOIE0);
392   return true;
393 }

```

The second memory bank (bank 1) memory is laid out manually. I use this space or cached elements.
Anything that needs to be accessed quickly I will aim to keep in (bank 0)

Edited 2nd time: I have this bad habit of using the word heap, in place of stack. fixed.

avr/cores/megacommand/MidiUart.cpp
```diff
@@ -277,14 +277,12 @@ bool MidiUartClass::avail() { return !rxRb.isEmpty(); }
 uint8_t MidiUartClass::m_getc() { return rxRb.get(); }
 
 ISR(USART1_RX_vect) {
-  uint8_t old_ram_bank = switch_ram_bank(0);
+  select_bank(0);
   isr_midi();
-  switch_ram_bank(old_ram_bank);
```
jmamma: The ISR functions can be called at any time during program execution, this includes during memcpy_bank1()

If we were in BANK 1 before the call to ISR(), then we must switch back to BANK1 when exiting the ISR.

avr/cores/megacommand/MidiUart.cpp
```diff
@@ -277,14 +277,12 @@ bool MidiUartClass::avail() { return !rxRb.isEmpty(); }
...
 }
 ISR(USART2_RX_vect) {
-  uint8_t old_ram_bank = switch_ram_bank(0);
+  select_bank(0);
   isr_midi();
-  switch_ram_bank(old_ram_bank);
```
jmamma: Same issue as above. must switch to old bank upon exit of ISR().

avr/cores/megacommand/MidiUart.cpp
```diff
@@ -456,11 +454,10 @@ ISR(USART1_UDRE_vect) {
   if (MidiUart.txRb.isEmpty()) {
     CLEAR_BIT(UCSR1B, UDRIE1);
   }
-  switch_ram_bank(old_ram_bank);
```
jmamma: Same issue as above. must switch to old bank upon exit of ISR().

avr/cores/megacommand/MidiUart.cpp
```diff
@@ -475,7 +472,6 @@ ISR(USART2_UDRE_vect) {
   if (MidiUart2.txRb.isEmpty()) {
     CLEAR_BIT(UCSR2B, UDRIE1);
   }
-  switch_ram_bank(old_ram_bank);
```
jmamma: Same issue as above. must switch to old bank upon exit of ISR().

avr/cores/megacommand/WProgram.h
```diff
@@ -88,4 +88,25 @@ extern inline uint8_t switch_ram_bank(uint8_t x) {
...
+#define select_bank(x) RamBankSelector __bank_selector(x)
+
+#endif// __cplusplus
+
+#define memcpy_bank1(dst, src, len) \
+  do { \
+  switch_ram_bank(1); \
+  memcpy(dst, src, len); \
+  switch_ram_bank(0); \
+  } while (false)
```
jmamma: what's the significance of "do { } while (false)" here? 🤔

avr/cores/megacommand/main.cpp
```diff
@@ -220,7 +220,6 @@ ISR(TIMER1_COMPA_vect) {
     }
   }
   }
-  switch_ram_bank(old_ram_bank);
```
jmamma: Same issue as above. must switch to old bank upon exit of ISR().

avr/cores/megacommand/main.cpp
```diff
@@ -277,7 +276,6 @@ ISR(TIMER2_COMPA_vect, ISR_NOBLOCK) {
   gui_poll();
 #endif
 
-  switch_ram_bank(old_ram_bank);
```
jmamma: Same issue as above. must switch to old bank upon exit of ISR().

jmamma (changes requested):
Generally I think all new commits should be pushed to DEV branch.
Only merge in to master after stability testing has been performed.

avr/cores/megacommand/WProgram.h
yatli: This is mainly a guard trick.
Without this, you could write:


if (something)
  memcpy_bank1(...)


when expanded, only the first statement will go into the conditional branch.

jmamma:
Revised my memory explanation for a second time. I mistakenly used the word heap in place of stack, which would be rather confusing.

yatli:
Thanks for explaining about the memory layout! Yeah I get the idea that starting from 0x2200 are selectable banks, I’ve seen this before in sram test (that’s why marking the pointers volatile is a good idea, similar to PROGMEM segment).

wrt the select_bank macro, it establishes a guard instance RamBankSelector, which restores old bank during destruction. This way it ensures that bank is restored when the current stack frame is popped(and the local instance goes out of life cycle) — this is pretty common for modern c++ (e.g. std::lock_guard) as RAII pattern. The benefit is that, for example, during ISR, you could return midway without worrying about manual bank restoration— the destructor does this.

Anyway, I do think this is a trade off here, because it hides bank selection details, and sometimes it may not be the best idea.

yatli:
Another question: I see the difference of MDTrack vs. MDTrackLight;

Only MDTrackLight is placed in bank1, the remainder, the PLocks, stay in the “heap” (yeah I get this) — is that part a shared data structure by 16 tracks?

jmamma:
Regarding MDTrack and MDTrackLight.

"is that part a shared data structure by 16 tracks?"

In Chain mode. MDTrack objects are read from the SDCard and cached in to bank2.
Chain mode does not use the MD's sequencer information.

MDTrackLight Class was used to create a smaller object with only the MC's internal sequencer data.

The larger MDTrack data structure is used for Write/Save functionality. An MDTrack object includes all the data for a MD track including some Kit settings. This data structure is unique per track, and none of the  memory elements are shared.

jmamma:
"wrt the select_bank macro, it establishes a guard instance RamBankSelector, which restores old bank during destruction. This way it ensures that bank is restored when the current stack frame is popped(and the local instance goes out of life cycle) — this is pretty common for modern c++ (e.g. std::lock_guard) as RAII pattern. The benefit is that, for example, during ISR, you could return midway without worrying about manual bank restoration— the destructor does this."

Cool. As you would have noticed, I haven't employed destructors. But this seems like an excellent application and might provide some additional flexibility to ISR()s in the future.

I will review the rest of your changes later tonight/tomorrow.

Thanks for the contribution. If anything, the code needs to be aligned with best practices. 
My time is limited (and sometimes i much rather be making music than dev), so help here is much appreciated.

yatli:
Ok that checks out. The extra data is not cached in the high bank.

If my memory layout specs hold, then it’s possible to pack in a few more tracks right, I’d like to get A4 FX/CV tracks for controlling a MS-20 mini.

jmamma:
Expanding to 6 x A4 Tracks should be do-able.

You need to decide if the CV + FX settings are important, and should they be stored alongside the sequencer data. ( I think they should ).

I haven't looked at the A4 sysex in a while, If i recall, CV + FX settings will be different from the sound spec.

Right now we don't do any actual decoding of the A4 data. We just store and send the sound dumps as they are. If you want to build a proper A4 pattern + kit + sound decoder then be my guest :-)

I know there may be some stack memory issues with write_mode as defined in MCLActions.
See the poorly named "MCLActions::send_pattern_kit_to_md()"

Basically, the A4 sounds are cached in the stack, and then sent to the A4 all in one hit. 
We may need to move this stack memory to bank2, to fit 6 tracks in.

---

On a more broader topic. I had always planned for tracks higher than 16, to be configurable as general MIDI sequencer tracks. The functionality is 90% there. Though the sequencer pages need to be modified to support entering note data in using the encoders.

avr/cores/megacommand/WProgram.h
jmamma: I'm wondering what the performance overhead for destructor in ISR is.

You'll notice that I inline within the ISR routines wherever possible to avoid switching the current stack.

avr/cores/megacommand/WProgram.h
```diff
@@ -88,4 +88,25 @@ extern inline uint8_t switch_ram_bank(uint8_t x) {
...
+  public:
+  RamBankSelector(uint8_t bank) { m_oldbank = switch_ram_bank(bank); }
+  ~RamBankSelector() { switch_ram_bank(m_oldbank); }
+};
+
+#define select_bank(x) RamBankSelector __bank_selector(x)
+
+#endif// __cplusplus
+
+#define memcpy_bank1(dst, src, len) \
```
jmamma: Could we inline memcpy_bank1, and avoid the while.

I feel a little uneasy about this trick.

avr/cores/megacommand/MCL/GridRowHeader.h
jmamma: I can't think of a scenario where you'd want c_GridWidth != GRID_WIDTH.
Thoughts?

Ideally GRID_WIDTH should be defined once. However, I am aware of an issue of compile/link fail because cannot find the definition of GRID_WIDTH.

One extra complication is that adding additional variables to the start of GridRowHeader will make reads to existing projects (created with the old GridRowHeader object) incompatible.

 I'm trying to limit ProjectSchema changes where possible for end-user quality of life.

yatli:
> You need to decide if the CV + FX settings are important, and should they be stored alongside the sequencer data. ( I think they should ).

I think they should. And my guess is that they share the same data structure with normal tracks -- substituting Osc1 with CV-A, Osc2 with CV-B, Filters with CV-C and Amp with CV-D.. Let me dig around here.

The CV machines have different options (other than the parameters, options are accessed with Fn+page) though -- CV control type (trig/gate etc.), polarity, voltage (up to 10v!). Gotta be careful 0.0

yatli:
btw I've rebased this PR upon `dev` and I'm done with the memory-related changes.
I created the `Bank1Object` class so that the bank1 accessors can be further unified.

avr/cores/megacommand/WProgram.h
yatli: most probably, the content of the destructor will be pulled out and inlined. this is a kind of zero-cost abstraction.

yatli:
In `MidiUart.cpp`:

```c++
#include <MidiClock.h>
MidiUartClass MidiUart;
MidiUartClass2 MidiUart2;
```

In `main.cpp`:

```c++
uint8_t sysexBuf[5500];
MidiClass Midi(&MidiUart, sysexBuf, sizeof(sysexBuf));
uint8_t sysexBuf2[2800];
MidiClass Midi2(&MidiUart2, sysexBuf2, sizeof(sysexBuf2));
```

In `platform.txt`:

```
compiler.cpp.flags=-c -g -Os {compiler.warning_flags} -std=gnu++11 -fpermissive -fno-exceptions -ffunction-sections -fdata-sections -fno-threadsafe-statics -MMD -flto -Wl,--gc-section,--defsym=__stack=0x8020ff,--section-start,.data=0x82100,--defsym=__heap_end=0x80ffff -I{build.core.path}/Midi -I{build.core.path}/GUI -I{build.core.path}/CommonTools -I{build.core.path}/Elektron -I{build.core.path}/MD -I{build.core.path}/MidiTools -I{build.core.path}/Sequencer -I{build.core.path}/MNM -I{build.core.path}/PageTools -I{build.core.path}/MDSketches -I{build.core.path}/MDPages -I{build.core.path}/A4 -I{build.core.path}/SPI -I{build.core.path}/Wire -I{build.core.path}/Adafruit-GFX-Library -I{build.core.path}/Adafruit_SSD1305_Library -I{build.core.path}/SdFat -I{build.core.path}/SDCard -I{build.core.path}/MCL
```

Ok I get this. We're telling the compiler "it's okay, relax, and put things beyond 0x2200" -- super cool!

jmamma (changes requested):

jmamma (approved):
Approving for now, will inline in seperate commit.

jmamma:
In the interest of time, i've got ahead and merge your commits to dev branch.

I've reverted the changes to GridRowHeader.*  (The duplicate declaration of GRID_WIDTH is a bug and needs to be fixed next time the project spec changes in master)

The changes to bank switching (using object destructor) need to be tested.

avr/cores/megacommand/WProgram.h
yatli: Sure. Let’s go c++ :)

jmamma:
In main.cpp we use  ISR_NOBLOCK for timer2. This prevents missed reads of the UART buffers, by re-enabling global interrupts inside the ISR.

If ISR(TIMER2_COMPA_vect) is interrupted after select_bank(0), will the destructor   ~RamBankSelector() be called?

yatli:
It won’t be called on interruption, because an interruption is effectively the same as a “thread being swapped out in the OS”, and does not change the current instruction pointer(when later coming back), and thus not changing the life scope of locals.

The destructor will be triggered after restoration, when the (first) ISR exits.

edit:

The destructor mechanism works by letting the compiler insert destruction code at every possible location that the thing goes out of scope. Think about an automated code assistant filling switch_ram_bank(old) for you when necessary.

jmamma:
Excellent. Thanks.

yatli:
Another thing I slightly worry about the memory layout is the alignment issue.

In a beefier machine where the words are wider, the compiler would leave paddings between members of a class, and thus making the size larger than the sum of the members.

Not sure what’s the case with avr-gcc and the 8bit mcu

yatli:
Reference: https://www.avrfreaks.net/forum/packed-struct

The potential problems are mitigated by adding __attribute__((packed)) before the class keyword.

jmamma:
Right. My assumption was that avr was not padding. For the classes i've checked manually in the past, sizeof was expected.

But I could be wrong here.

Are you concerned with wasted RAM?

yatli:
Nope, I worry more about overlapped tracks. That’ll be a problem.

Yatao

________________________________

jmamma:
My understanding was that sizeof() should return correct object size, including any potential padding. 

Given that we use len = sizeof() this should mitigate collisions ?

yatli:
I can take care of this. sizeof gives faithful results, but I’d need to adjust Bank1Object a bit.

Yatao

________________________________

yatli:
Bank1Obj is updated. This mixin is now able to access the true size of the attached class.

jmamma (approved):

PR #5, opened 2019-03-26
sysex reveng. on A4

yatli:
this is a work-in-progress.
uploading to share initial findings...


yatli:
uh oh I've sent 0x7C, 0x7D.

Measuring CV out now...

yatli:
CV out looks good, drives ms20 well.. doesn’t look like it’s a problem after the “calibration”

jmamma:
These are the references I know of:

https://github.com/bsp2/libanalogrytm

[mm1_22_sysex_0_6.pdf](https://github.com/jmamma/MIDICtrl20_MegaCommand/files/3007499/mm1_22_sysex_0_6.pdf)

[md1_53-sysex-0_8.pdf](https://github.com/jmamma/MIDICtrl20_MegaCommand/files/3007511/md1_53-sysex-0_8.pdf)

https://www.elektronauts.com/t/decoding-the-digitone-sysex/

It would be great if the encoding / decoding functionality could be made consistent across devices. This would either mean using Wesen's original implementation, or creating a new implementation and back porting to the MD + MNM

yatli:
MD command 0x40 is amazing.. I’m wondering if/how we could find that on other models..

yatli:
enough banging.. collector's up. time to do some real job

I'm thinking about bringing in some formalism. For example with theorem provers, I can raise theories about the protocols, and the prover will try to prove me wrong, given the constraints and a database of samples.

yatli:
in case the prover is satisfied, we will have a working model of the protocol. the model may even contain parameters that are symbolically deduced from samples rather than given by hand..

yatli:
progress:

got a nice little protocol monitor;
did a quick and dirty session and found that trigger bits and note pitch info are recorded separately.

Trigger data is stored at the beginning:
image

-- okay, one bit effective, among a few unknown bits :p

And notes later, at 0x9A:
image

Those `7F`s are fillers for inactive triggers.
`3C` is note `C4`
`48` is note `C5` -- standard midi

jmamma:
Hi Yatao,

Can you make note of the size of A4 SYSEX dumps somewhere.

Thanks,

- Justin

yatli:
@jmamma the response sizes are attached to `A4Cmds.fs`.

yatli:
pattern & sound formats are partially decoded. see `Hypothesis.fs`

formal methods are overkill -- nothing fancy going on here.

proto_analyzer/Hypothesis.fs
```diff
@@ -0,0 +1,45 @@
...
+
+// TODO pattern note pitch values are standard midi notes
+// TODO pattern dump offset 0x216: normal pattern len
+// TODO pattern dump offset 0x20f: normal pattern spd, 0x30 is 1/8, 0x3C is 1x
+// TODO tracks are store sequentially in a pattern - Track1-4, then FX, then CV;
+// TODO every track can have its own pattern scale settings.
+// TODO for both trigger and note pitch data, payloads are grouped in 7-byte blocks, separated by flipping 2a 55 2a 55s (acting as sync bytes?)
+//      00 in place of 2a/55 means no data in the block
+//      so, the raw bytes:  01 40 09 55 41 01 44 09 45 01 48 2a 09 49 01 4c 09 4d 01 55 50 09 51 01 54 09 55 2a
+//      are really:         01 40 09    41 01 44 09 45 01 48    09 49 01 4c 09 4d 01    50 09 51 01 54 09 55
```
jmamma, line 11: This sounds like elektron's 7bit encoding method. To pack an array of bytes in to an array of 7 bit values, they take 7 bytes at a time. Each byte has the MSB removed, this MSB is stored in an extra byte for each of the 7 values.

So your 7bit data stream looks like:
[ MSBs ]  [ 7bitValue ]  [ 7bitValue ]   [ 7bitValue ]   [ 7bitValue ]   [ 7bitValue ]   [ 7bitValue ]   [ 7bitValue ]  (repeat)

"How the 7 bit encoding is generated
As a SysEx message only allows 7 bit data, 8 bit data blocks are converted as follows:
The first data (byte) contains the MSB of the following, up to, 7 data, data[0..6]. Bit 6 is the MSB bit of data[0] and bit 0 is the MSB of data[6]. 7 bytes encodes to 8 data. 8 bytes encodes to 10 data."

jmamma:
> pattern & sound formats are partially decoded. see `Hypothesis.fs`
> 
> formal methods are overkill -- nothing fancy going on here.

great work so far.

proto_analyzer/Hypothesis.fs
jmamma: I should add. Please refer to the mnm pdf i uploaded explaining packing and encoding.

yatli: thanks, that's very interesting. I'll read up the mnm manual to follow up.
the MSBs probably aren't used here though, as the truth table already covers all the information for triggers.
also, please note:
```
0x55 = 0101 0101
0x2a = 0010 1010
```
So they add up to 0x7F (and do not chance however I arm the triggers) -- more like a synchronization pattern -- in case some bytes are dropped, it still recovers "on the beat".
My reasoning goes the same with the 01-09 pattern.
Another explaination would be, that all the effective bytes (0x40, 0x41 etc) have the MSB toggled, and the 01-09 patterns do not. If so, because each block contains 7 bytes, the pattern naturally shuffles like `1010101   0101010  1010101  0101010 ... `.

it surprised me that the 7bit and 8bit COM protocols are compatible, so the monitor program works without even touching the width and stop bit settings. :p

yatli:
it's a bit boring to make 100 patches to identify the params but there we have it :)
patch 0 is the reference setting, and each of the following have one param tweaked.
the order is noted in `A4Types.fs`, namely `type Sound = { ... }` record type.

the patches are extracted and added to `samples.7z`.

there's one part at the end of each patch that I don't understand yet (offset 0x19a-0x19b):

image

these two bytes keep changing on all the patches. could be a checksum of some kind. does not look like a shuffling scheme though, because the values are so concentrated around 0x3d00-0x3f00, and more tweaks in the patch = higher checksum.

might be just a direct sum over the bytes 0.0

jmamma:
See the MonoMachine pdf for how checksums work:

```
How the checksums are calculated
The checksum is generated from the data marked in blue in the tables. It is the lower 14 bits of the sum of all the indicated data.

[0 ... 127] Checksum (bit 7...13)
[0 ... 127] Checksum (bit 0...6)
[0 ... 127] Message length (bit 7...13)
[0 ... 127] Message length (bit 0...6)
0xF7 End of SysEx
```

yatli:
great! with that piece of information installed, I can start writing a c++ prototype for sound patches very soon.

yatli:
update:

working on the prototype now; please see `A4Messages{.h,.cpp}`

also, the analyzer is giving me positive results. please see `sounds.inference.txt`

yatli:
have to say, this is a bit convoluted:

```c++
  decoder.skip(1); // skip sound header 0x05
  decoder.get(tags);
  decoder.get(name);
  decoder.get(osc[0].tuning);
  decoder.get(osc[0].fine);
  decoder.get(osc[1].tuning);
  decoder.get(osc[1].fine);
  decoder.get(osc[0].detune);
  decoder.get(osc[1].detune);
  decoder.get(osc[0].keytrack);
  decoder.get(osc[1].keytrack);
```

is it ok to use the decoder like this?
from what you're doing in the other PR, I guess batch reads are more effective.

jmamma:
> update:
> 
> working on the prototype now; please see `A4Messages{.h,.cpp}`
> 
> also, the analyzer is giving me positive results. please see `sounds.inference.txt`

Just had quick glance. Amazing. looking good so far!

jmamma:
> have to say, this is a bit convoluted:
> 
> ```c++
>   decoder.skip(1); // skip sound header 0x05
>   decoder.get(tags);
>   decoder.get(name);
>   decoder.get(osc[0].tuning);
>   decoder.get(osc[0].fine);
>   decoder.get(osc[1].tuning);
>   decoder.get(osc[1].fine);
>   decoder.get(osc[0].detune);
>   decoder.get(osc[1].detune);
>   decoder.get(osc[0].keytrack);
>   decoder.get(osc[1].keytrack);
> ```
> 
> is it ok to use the decoder like this?
> from what you're doing in the other PR, I guess batch reads are more effective.

The decoder code is all of Wesen's original work. Batch reads should be okay and will be more efficient, and reduce program size.

A long, long time ago I fixed a bug MDPattern.cpp relating to batch reads.
 Looking at the code now, I can't actually work out why it failed, but the pattern trig masks would end up incomplete.

Original code (https://github.com/wesen/mididuino/blob/master/hardware/libraries/MD/MDPattern.cpp):
```
  decoder.start7Bit();
  decoder.get32(lockPatterns, 16);

  decoder.start7Bit();
  decoder.get32(&accentPattern, 4);
  /*
    accentPattern = decoder.gget32();
    slidePattern  = decoder.gget32();
    swingPattern  = decoder.gget32();
    swingAmount   = decoder.gget32();
  */
  decoder.stop7Bit();
```

Patched code (MDPattern.cpp):

```
111     decoder.start7Bit();
112     decoder.get32(lockPatterns, 16);
113     
114     decoder.start7Bit();
115     
116     decoder.get32(&accentPattern);
117 
118     decoder.get32(&slidePattern);
119     decoder.get32(&swingPattern);
120 
121     decoder.get32(&swingAmount);
122 
123     /*
124      accentPattern = decoder.gget32();
125      slidePattern  = decoder.gget32();
126      swingPattern  = decoder.gget32();
127      swingAmount   = decoder.gget32();
128      */
129     decoder.stop7Bit();
```

yatli:
Hmm, I suspect it's call start7Bit() twice? The state may be lagging behind and the decoder ate some bytes.

yatli:
I've cleaned up the inference results.
there were noisy bits possibly due to encoder sensitivity (reset to 0, then bounced to 1)

I'm going to figure out how to efficiently decode them.

yatli:
@jmamma I've rolled the parameters into a big struct. that should make the codec process more efficient. I think the codec is up for some testing!

jmamma:
Amazing.

> @jmamma I've rolled the parameters into a big struct. that should make the codec process more efficient. I think the codec is up for some testing!

I think this will be okay for A4Sound. I was thinking yesterday that A4Pattern at 14KB is going to be too big to keep around as global object. We may need to break it down in to loadable sections.

yatli:
> I was thinking yesterday that A4Pattern at 14KB is going to be too big to keep around as global object. We may need to break it down in to loadable sections.


they're not packing things as tight as possible. we can squeeze ~40% out without compression.
with compression I'm not sure. but maybe not worth it.

can we at least fit it into the sysex recv buffer?

yatli:
another approach, if we can afford decoding just-in-time while sysex is receiving, then we decode the tracks one by one reducing the sysex size to 25%

jmamma:
> they're not packing things as tight as possible. we can squeeze ~40% out without compression.
> with compression I'm not sure. but maybe not worth it.
> 
> can we at least fit it into the sysex recv buffer?

The pull request (sysex refactor) that I'm working on is to move to sysex buffers to the HIGH memory bank. This has two benefits, it will allow us to process much large sysex dumps, and secondly free's up more main memory for timing critical code, like the sequencer.

yatli:
ah ok, no need to do JIT decoding. leave more space for the sequencer is the right way to go.

yatli:
While you're on the sysex infrastructure, I will work out A4kit/A4pattern codec. :)
Let's see how much space we can squeeze from it.

jmamma:
I just got the ArduinoIDE to compile with new sysex code and we have 
17170 Bytes of RAM free.

Not bad :) .

jmamma:
Finally had time to look through the code. Great work here.

jmamma:
Thought I had already merged this one..

jmamma:
```/Applications/Arduino.app/Contents/Java/hardware/valence/avr/cores/megacommand/A4/A4Messages.hh:8:21: fatal error: algorithm: No such file or directory
 #include <algorithm>
                     ^
````

Getting compile error

**Comment by jmamma** [comment]
@yatli Do we need to include a c++ STL?

Were you using something like ?

https://github.com/maniacbug/StandardCplusplus

**Comment by jmamma** [comment]
```
   a4sfloat_t(uint8_t dat[2]) { std::copy(dat, dat + 2, data); }
```

I think we should avoid using the STL, due to increase in program size.

For example, the implementation above could be done using memcpy, which is faster than a while loop routine with pointers.

**Comment by yatli** [comment]
yeah I was in a STL mood back then but it's obviously not ideal to link against libc++!

**Comment by jmamma** [comment]
I removed the reference to C++ STL.

I realised that the A4Sound class has changed. Which could possibly conflict with the byte layout of the old A4Sound, leading to incompatibility ?

Also A4Sound::fromSysex(MidiClass *midi) { } needs to be updated.

**Comment by jmamma** [comment]
This now compiles , but saving and sending sounds from MCL does not appear to be working correctly.
Need to check fromSysex()

**Comment by jmamma** [comment]
New A4 firmware. 1.40 released.

Need to check compatibility at some point.

@yatli

**Comment by yatli** [comment]
Wow, conflicts.
Needs a good deep dive into this thread after the MD side.

**Comment by yatli** [comment]
@jmamma this branch is up-to-date with `dev`. I can resume the breakpoint now.

**Comment by yatli** [comment]
good news. A4 recognizes the transmission as sound patch!

now just have to fix the incompatibility between old blob format and the new struct.

File: avr/cores/megacommand/A4/A4Messages.cpp [hunk 2]
```diff
@@ -58,7 +58,7 @@ uint16_t A4Global::toSysex(ElektronDataToSysexEncoder &encoder) {
 static constexpr uint8_t a4sound_prologue[8] =  { 0x00, 0x20, 0x3c, 0x06, 0x00, 0x53, 0x01, 0x01 };
 static constexpr uint8_t a4soundx_prologue[8] = { 0x00, 0x20, 0x3c, 0x06, 0x00, 0x59, 0x01, 0x01 };
 static constexpr uint8_t a4sound_header[8] = { 0x78, 0x3e, 0x6f, 0x3a, 0x3a, 0x00, 0x00, 0x00 };
-static constexpr uint8_t a4sound_footer[4] =  { 0x3a, 0xe3, 0x7a, 0x4e };
+static constexpr uint8_t a4sound_footer[4] =  { 0x3a, 0x3e, 0x7a, 0x4e };
```
yatli, line 61: #FAIL

yatli:
@jmamma we should decide whether to keep the compatibility with the current A4 patches.

yatli:
Tested, mixed result. Some patches can be loaded properly, others result in a "blank" patch, everything stuck at leftmost position (-64, 0, off, etc)

yatli:
I've written a simple predicate to test whether a patch is legacy blob or structured.
See A4Sound::is_legacy_format()

jmamma:
> @jmamma we should decide whether to keep the compatibility with the current A4 patches.

What's involved to do this? Maintain two classes?

I upgraded A4 firmware a few weeks ago, hadn't noticed sound dump incompatibility. But my testing was not thorough.

yatli:
The sysex dumps from A4 haven't changed. I've replaced the blob payload in `A4Sound` with structured parameters, and that strips out the 7bit encoding.

yatli:
The hack:

```C++

  // === Begin new a4 sound patch layout
  uint8_t          tags[4];      // 32 tags
  char             name[16];     // null-terminated
  // sizeof(a4sound_t) is 318, which encodes 363 bytes in 7-bit enc, + 42 bytes header metadata + 8B prologue + 2B SYSEX frame = 415B
  a4sound_t        sound;

  // === End new a4 sound patch layout, 338B

  // old a4 sound patch layout: uint8_t[415 - 10 - 2 - 4 - 1], 398 bytes
  // -10 : prologue and 0x78 0x3E, part of a4 sound header.
  // -2: sysex frame
  // -4: len & checksum
  // -1: origposition
  uint8_t          sound_pad[60];

  bool is_legacy_format();
  void fromLegacySysexDump();

bool A4Sound::is_legacy_format() {
  uint8_t* payload = this->tags;
  return !memcmp(payload, a4sound_header + 2, 6);
}

void A4Sound::fromLegacySysexDump() {
  toggleLed2();
  // legacy payload len = 398
  uint8_t* payload = this->tags;
  // skip the partial header 
  payload += 6;
  // getting data from itself. always a few steps behind.
  ElektronSysexDecoder decoder(DATA_ENCODER_INIT(payload, 392));
  fromSysex_impl(decoder);
}
```

jmamma:
Ah yeah. Okay. 

This is probably a more broader issue. Do we support legacy project formats ?

For instance, there are changes id like to make the MDTrack structure, but i've been holding off because it will break all my projects.

yatli:
Is the project format versioned? If so we can perform one-way upgrade on it without maintaining different versions of actual working classes.

jmamma:
I'm thinking we stick with legacy sound dump until version 3.0

Major version change should == project spec change.

jmamma:
> Is the project format versioned? If so we can perform one-way upgrade on it without maintaining different versions of actual working classes.

Yes it is. 

If you can think of a lightway way to achieve this, I'd be happy to convert projects on load.

jmamma (changes requested):
You mean DEVICE_MNM ?

avr/cores/megacommand/Midi/MidiID.hh
```diff
@@ -7,6 +7,7 @@
 #define DEVICE_NULL 0xFF
 #define DEVICE_MIDI 0xFE
 #define DEVICE_MD 0x02
+#define DEVICE_MD 0x03
```
jmamma: DEVICE_MNM ?

jmamma: Also FYI MidiID does not work on the new machines last time I checked.

yatli: If so we can use a MNM-specific sysex query as a probe.
Newer machine=digitakt etc?

yatli:
found a better home for this branch.

yatli:
Ready for project format conversion. No custom version probe needed this time.

PR #12, opened 2019-04-06
Ring buffer refactor

jmamma:
@yatli refactord the MIDI ring buffer code to use membank1. 

We now have 23371Bytes of RAM available.

You might have a better solution to modifying the template to use ptr, instead of creating an array of size N (buf[n])

My implementation is a bit of a hack. I create ring buffers of size 0. then assign pointer and length in the MidiUart constructors.


jmamma:
Looks like there's a problem with buffer sizes.

I miscalculated. earlier the original buffers were of type int16_t
so double the size.

yatli:
I'm thinking about this. Now we have quite a few `if (ptr == NULL) { .. do with unknown ptr .. } else { .. do with bank1 }` branches. These can be unified, and, their behavior can be statically deduced because we assign them at compile time anyway.

jmamma:
This pull request is still not stable. Getting crash after some time of use.

avr/cores/megacommand/memory.h
```diff
@@ -47,21 +47,18 @@ class RamBankSelector {
 #endif// __cplusplus
 
 extern inline void memcpy_bank1(volatile void *dst, volatile void *src, uint32_t len) {
-  switch_ram_bank(1);
+  select_bank(1);
```
yatli, line 84: I thought these should be equivalent? 🤔

jmamma: Because the ring buffer code is being called within UART ISR, there may be some circumstances where the original bank, before the ISR routine was already bank1.

So switching back to bank 0 will cause things to blow up when the program exits the ISR.

jmamma: There's still something wrong with this code. I was playing around with it late last night.
I'm not sure if it's incorrect bank switching, or memory leak, or something else.

avr/cores/megacommand/CommonTools/RingBuffer.h
```diff
@@ -104,8 +109,14 @@ template <class C, int N, class T>
     overflow++;
     return false;
   }
+  if (ptr == NULL) {
   memcpy((void *)&buf[wr], (void *)c, sizeof(*c));
-  wr = RB_INC(wr);
+  }
+  else {
+  memcpy_bank1(ptr + wr, (void *)c, sizeof(*c));
```
yatli: The behavior differs from the original.
`ptr` is `volatile uint8_t*`, but `buf` is `C[N]`.
`&buf[wr]` is `buf + sizeof(C) * wr`.

avr/cores/megacommand/memory.h
yatli: oww, I got it; switch_ram_bank(0) wasn't ideal.

yatli (changes requested):

yatli:
@jmamma anything using a ringbuf where `sizeof(C) != 1`?

avr/cores/megacommand/memory.h
yatli: going over the code.

avr/cores/megacommand/memory.h
```diff
@@ -1,7 +1,22 @@
 #ifndef MEMORY_H__
 #define MEMORY_H__
 
-#define BANK1_SYSEX1_DATA_START 0x2200
+#define RX_BUF_SIZE 0x80
+#define TX_BUF_SIZE 0x0C00
```
yatli: gotcha.
TX_BUF_SIZE is `0x0C00` so the buflen type should be `uint16_t` instead of default `uint8_t`.

jmamma:
> @jmamma anything using a ringbuf where `sizeof(C) != 1`?

Yes.

See : GUI/Events.hh

avr/cores/megacommand/memory.h
yatli: I've seen the `#if (TX_BUF_SIZE >= 256)` part. not an issue.

yatli (changes requested):
https://github.com/jmamma/MIDICtrl20_MegaCommand/blob/8420cff8f747e597e6a03963ba4c40c92b72fea7/avr/cores/megacommand/CommonTools/RingBuffer.h#L132

There are remaining parts not multiplexed over `N` and injected `len`.

yatli:
I'll get back my free time after 24th (and recover from the burnout).
If time permits, I'll take this one?

yatli:
@jmamma 
I merged `master`, and now the compiler reports:

```
Sketch uses 20796 bytes (8%) of program storage space. Maximum is 253952 bytes.
Global variables use 3201 bytes (5%) of dynamic memory, leaving 53630 bytes for local variables. Maximum is 56831 bytes.
```

is it magic or my configuration is wrong?

yatli:
nevermind, it's a #ifdef MCL directive masking out the real meat.

jmamma:
> nevermind, it's a #ifdef MCL directive masking out the real meat.

:smile: I fell in to this trap on the weekend. scratched my head for a few minutes.

yatli:
I've downloaded this branch to MC and played for a few minutes.
Looks like there's still some problem: at a time point the connection seems to be broken. after that, MC responses slowly, and chain operations fail. It cannot download patterns from MD, not even switching to channel 16 to leverage the TI.

yatli:
if it works fine on `master` (I haven't tried yet) then it must be related to the changes in this PR. I'll investigate this tonight.

jmamma:
Yes, this PR is unstable.

Need to investigate what the cause is. Perhaps the buffers are unexpectedly filling up, and causing weird behaviour, You should be able to check this by adding some if statements and print len in get and put.

The other potential cause is an issue with bank switching, or some type of memory overflow ?

yatli:
great. I'll go over the code more carefully. there're also clues left by the 1 month younger me.

jmamma:
It appears txRb.isFull()) condition is being triggered?

You can see LED2 set to HIGH.

```
216 void MidiUartClass::m_putc(uint8_t c) {
217   //#ifdef BLAH
218   if (c == 0xF0) {
219     uart_block = 1;
220   }
221   if (c == 0xF7) {
222     uart_block = 0;
223   }
224 again:
225 
226   if (txRb.isFull()) {
227     while (txRb.isFull()) {
228       if (IN_IRQ()) {
229         // if we are in an irq, we need to do the sending ourselves as the TX
230         // irq is blocked
231         setLed2();
```

yatli:
trying this led trick

jmamma:
think i found the problem...

yatli:
@jmamma which one is `Led2`, on the left?

jmamma:
led2 is right

jmamma:
in MidiUart.h
```
 75 
 76   volatile RingBuffer<0, RX_BUF_TYPE> rxRb;
 77 
 78 #ifdef TX_IRQ
 79   volatile RingBuffer<0, TX_BUF_TYPE> txRb;
 80 #endif
```

I'm setting N = 0, to indicate that we're using membank2. 
However the ringbuffer code uses N, in stead of len:


For example:
```
@@ -129,7 +129,7 @@ template <class C, int N, class T>
   if (ptr == NULL) { ret = buf[rd]; }
   else { ret = get_byte_bank1(ptr + rd); }
   rd++;
-  if (rd == N) { rd = 0; }
+  if (rd == len) { rd = 0; }
   return ret;
 }
 ```

**Comment by yatli** [comment]
still crashing.

**Comment by yatli** [comment]
I'm still trying to figure out what I meant with the comments left. Trying to sync up :)

**Comment by jmamma** [comment]
Did you clean before make ?

---

I'm testing new patch. Crashed at first, but I realised RingBuffer.h was not recompiled.

**Comment by yatli** [comment]
```
The behavior differs from the original.
ptr is volatile uint8_t*, but buf is C[N].
&buf[wr] is buf + sizeof(C) * wr.
```
ok I got what I meant. Fixing this one.

**Comment by yatli** [comment]
whoa, there was a copy/paste error -- should be rd but we put wr there

**Comment by jmamma** [comment]
Ha, I was looking at that as well. 

> whoa, there was a copy/paste error -- should be rd but we put wr there

where?

**Comment by yatli** [comment]
```
@@ -141,7 +141,7 @@ template <class C, int N, class T>
   memcpy(dst, (void *)&buf[rd], sizeof(C));
   }
   else {
-  memcpy_bank1(dst, ptr + wr, sizeof(C));
+  memcpy_bank1(dst, (void*)&static_cast<volatile C*>(ptr)[rd], sizeof(C));
   }
```

was reading from the wr head

**Comment by jmamma** [comment]
git blame me.

**Comment by yatli** [comment]
in general I think we should use `volatile C* ptr`, not `uint8_t`

**Comment by jmamma** [comment]
> in general I think we should use `volatile T* ptr`, not `uint8_t`

That's fine by me. 

MidiUart.cpp will need to modified.
```
 12 MidiUartClass MidiUart((volatile uint8_t *)BANK1_UART1_RX_BUFFER_START,UART1_RX_BUFFER_LEN,(volatile uint8_t *)BANK1_UART1_TX_BUFFER_START,UART1_TX_BUFFER_LEN);
 13 MidiUartClass2 MidiUart2((volatile uint8_t *)BANK1_UART2_RX_BUFFER_START,UART2_RX_BUFFER_LEN,(volatile uint8_t *)BANK1_UART2_TX_BUFFER_START,UART2_TX_BUFFER_LEN);

```

Replace  (uint8_t *) with RX_BUF_TYPE/TX_BUF_TYPE

I should think

**Comment by jmamma** [comment]
I have to head off for a few hours. 

My limited testing is showing the stability as resolved.

**Comment by yatli** [comment]
If they were using `uint8_t` then no need to change. I'm wrapping up compatible code.

**Comment by yatli** [comment]
have a nice one. I'll have some pressure test tonight.

File: build.ps1 [hunk 4]
```diff
@@ -0,0 +1,33 @@
+$pattern = "#pragma message:" 
+
+Write-Host "============> Build started."
+
+$buildOutput = &{
+    arduino compile --warnings default -b MIDICtrl20_MegaCommand:avr:mega .\sketch\
```
jmamma, line 6: Are you compiling here using the Arduino IDE?

I switched to Makefile because the Arduino IDE is extremely slow. 
Compiling source used to take around 5minutes. I can do it in under 20seconds with multi-threaded Make.

avr/cores/megacommand/MCL/MCLMemory.h
```diff
@@ -12,15 +12,36 @@
 
 //#define MD_TRACK_LEN (sizeof(GridTrack) + sizeof(MDSeqTrackData) + sizeof(MDMachine))
 //#define A4_TRACK_LEN (sizeof(GridTrack) + sizeof(ExtSeqTrackData) + sizeof(A4Sound))
+
+//Use these to produce compiler errors that probes the sizes!
+//template<int X> struct __WOW;
+//__WOW<sizeof(MDTrackLight)> szmd;
+//__WOW<sizeof(A4Track)> sza4;
+#pragma message("MD_TRACK_LEN = 501")
```
jmamma, line 20: Should this not be hardcoded, or have I misunderstood the intention here?

build.ps1
yatli: I'm using the arduino CLI, it's much better than the IDE: https://github.com/arduino/arduino-cli

avr/cores/megacommand/MCL/MCLMemory.h
yatli: during compile-time, sizeof will not calculate the actual value in a macro. To get it at compile time, uncomment the templated probe `__WOW`, but it produces compiler errors to show the info :O

so if we want to update this value, we have to stop the pipeline, produce some errors, extract the value and hardcode it back here. anyway it's compile time stuff.

Issue #20, opened 2019-05-07
Receive of large SYSEX dump causes internal sequencer to go out of sync


jmamma:
When transmitting Pattern from MD -> MC , the internal sequencer drops out. 

If you want to save current Pattern in to the grid during a live performance, the sequencer will drop out unpredictably whilst the MD pattern is being transferred.

Solution would be to prevent MD Patterns from being transferred when sequencer is running. Though this may cause confusion to users who expected their MD trigs to be merged.

jmamma:
```I'm guessing the MD is unable to process MIDI note trigs when it's transferring pattern dump. kit dump doesnt appear to cause this problem.```

unconfirmed. could be interrupts on the MC being overwhelmed by the pattern dump.

**Comment by jmamma** [comment]
Problem appears to be on the MC side. If I remove MIDI in when sending pattern from MD, the internal sequencer does not drop out when MD is transmitting a pattern.

**Comment by jmamma** [comment]
Sysex receive can be on either port 1 or port 2. Internal sequencer tracks are going out of alignment.

Need to check if mod12 counter is going out of sync with mod6 counter,

**Comment by jmamma** [comment]
Found evidence that interrupts were being re-enabled during MidiClock callbacks.
MidiClock callbacks trigger the sequencer events.

Looking at the code, seems this is the case
```218   void MidiClockClass::callCallbacks() {
219     if (state != STARTED)
220       return;
221 
222     // Moved MidiClock callbacks to Main Loop
223 
224     static bool inCallback = false;
225     if (inCallback) {
226       return;
227     } else {
228       inCallback = true;
229     }
230 
231 #ifndef HOST_MIDIDUINO
232     sei();
233 #endif
```

Need to work out how to proceed from here. hmm

PR #21, opened 2019-05-08
Isr rework

jmamma:
Based on https://github.com/jmamma/MIDICtrl20_MegaCommand/issues/20

I decided to try simplifying the ISR code.

This seems to be the way forward.

The above patch does not suffer from sequencer drop out during sysex dump receive, but the sequencer timing seems swing when entering various pages. I'm suspecting this has to do with the transmit ISRs lagging.

### DIFF
--- a/avr/cores/megacommand/CommonTools/RingBuffer.h
+++ b/avr/cores/megacommand/CommonTools/RingBuffer.h
@@ -26,14 +26,13 @@
  * index variable of type uint8_t. If your ringbuffer has more than
  * 255 elements, use T = uint16_t.
  **/
-template <class C, int N, class T = uint8_t>
-class CRingBuffer {
-	/**
-	 * \addtogroup helpers_cringbuffer
-	 * @{
-	 **/
-  
- public:
+template <class C, int N, class T = uint8_t> class CRingBuffer {
+  /**
+   * \addtogroup helpers_cringbuffer
+   * @{
+   **/
+
+public:
   volatile T rd, wr;
   volatile C buf[N];
   volatile T len = N;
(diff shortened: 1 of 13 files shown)


jmamma:
Firmware performance is better with this patch compared to dev.

However, there appears to be a higher probability of sequencer drop outs and timing jitter when entering sub pages (subjective observations). 

Wonder if there is more tweaking that can be done here.

jmamma:
Transmit simplification seems to have eliminated swing.

Still getting occasional drop out when entering Pages that request sysex.

jmamma:
https://ww1.microchip.com/downloads/en/AppNotes/doc8453.pdf

Well i've learnt a few things today.

inline can be ignored by the compiler at it's own discretion.

We are using the avr-gcc -Os flag which means code is being optimised for size.

jmamma:
Played around with variations of this implementation.

Specifically looking at how the mcl_seq.seq() function (executed by MidiCallback on clock pulse) interacts with the ISR code.

Disabling sei() in MidiClock callback leads to inconsistent tempo calculation.
Which would suggest that mcl_seq.seq() is not exiting before next clock pulse.

jmamma:
Looks like f6c7674 was causing some serious slowdown in the ISR.

MIDI timing overall should be improved. 

I see more scope for optimisation, but things are feeling almost as stable as the old isr implementation.

jmamma:
might need to revert 
38892a9.

testing is still showing tempo-miscalc.

jmamma:
Celebrated too early. Need to retest above patches.

yatli:
looks like I'm missing out a lot again :) 
will catch up with you.

jmamma:
I've re-enabled global interrupts in the ClockCallback code. I've confirmed that there are no collisions between clock callback routines, so I think it's safe to do so.

This resolves the tempo calculation issue.

I still cant work out why the tempo calc is off, if I disable this.

jmamma:
Sequencer drop out is much improved when receiving sysex dumps.

However in the events in which it fails I see:

func_call: store_tracks_in_mem
clock collision
WRONG LENGTH
5400

jmamma:
Need to compare this pull request against the old isr code. 

Would be good to have some MIDI bench that could compare sequencer jitter and latency
@yatli

yatli:
I have been focusing on building a better toolchain for development and now I have https://github.com/yatli/fvim.

I'm going through the code and trying to figure it out.

jmamma:
The old tempo calculation which ran in ISR was causing sysex dumps to have bad length.
Scenario was, clock received on port 2, sysex dump received on port 1.

jmamma:
I discovered last night that both the new and old ISR code had issues receiving SYSEX dumps and transmitting sequencer information simultaneously.

The old ISR code prioritises receiving of UART data and thus the sequencer will often drop out if dumps are received when the clock is running.

The new ISR code attempts to resolve this by running the ISR code as quickly as possible.

In the new ISR code, I observed that whilst the sequencer drop out was less likely during a SYSEX dump receive, the dump might not be received correctly (truncated).  It looks like the tempo patch above has fixed truncation (at least for kit receive), pattern receive may still not be 100% reliable.

--

On a side note. 8x turbo is probably too much for the 16MHz processor.

```
31250 * 8 / 10 = 25,000 BPS.   [ 10 = 8 data bits, + 2 start/stop ]
16MHZ /( 31250 * 8 / 10) = 640 HZ.
```

During max data transfer (a SYSEX dump receive for example). 
We only have 640 instructions before we have to read the next byte.

jmamma:
More testings has shown truncation still occurs. 

Similar scenario:
clock collision
WRONG LENGTH

jmamma:
It's looking like we'll have to disable the sequencer during pattern receive. Or prevent "Save" from retrieving the current pattern if sequencer is running (this will mess up auto-merge).

Kit receive we can squeeze between 16th notes.

jmamma:
disable the sequencer during pattern receive. -> still getting some timing problems here.

I'm looking at disabling pattern_rec when sequencer is running. As write mode requires this functionality, I'm looking at deprecating write mode in next release.

jmamma:
Almost ready to merge this one in to dev.

Any objections?

jmamma:
Probably need to do more testing with both ports active.

PR #23, opened 2019-05-26
Remove write mode

jmamma:
It's time to say goodbye to write mode.

We will be keeping the basic kit/pattern write functionality in the background for when the sequencer is not running. For instance, using the chain page with sequencer stopped will cause the MD current pattern and kit to be regenerated from the MC.


jmamma:
Should auto-merge be activated by default -> no longer allowing MD patterns to be restored from the chain/write page.

Should we still try and support "poor man's plus drive".

I'm leaning towards no.

yatli:
>poor man’s plus drive

Let’s make it a menu item then — it’ll be even better than a plusdrive. Access snapshots from an sdcard—how awesome is that ;)

That being said, it’s a low priority thing. No need to put it on a near milestone.

jmamma:
Write mode now removed.

Just need to continue to test that saving/merging and chaining works as expected.

jmamma:
I've merged this branch in to dev.

Issue #24, opened 2019-05-28
Let's optimize IS_BIT_SET


jmamma:
In CommonTools/helpers.h

We have the following macros.
```
/** 64-bit macros. **/
#define SET_BIT64(port, bit)   ((port) |= (uint64_t)(1ULL << (bit)))
#define CLEAR_BIT64(port, bit) ((port) &= (uint64_t)~(1ULL << (bit)))
#define TOGGLE_BIT64(port, bit) ((port) ^= (uint64_t)(1ULL << (bit)))
#define IS_BIT_SET64(port, bit) (((port) & (uint64_t)(1ULL << (bit))) ? 1 : 0)
#define IS_BIT_CLEAR64(port, bit) (((port) & (uint64_t)(1ULL << (bit))) == 0 ? 1 : 0)
```
IS_BIT_SET64 is used heavily in the sequencer ISR routines.

According to avr forum, the 32bit variant is extremely inefficient.

https://www.avrfreaks.net/forum/fast-nth-bit-set-long-code

@yatli how are your bit manipulation skills?

We can use https://godbolt.org/ to compare instruction length of various implementations

yatli:
ha!
Challenge accepted.

yatli:
```C
#define SET_BIT64(port, bit)   ((port) |= (uint64_t)(1ULL << (bit)))
```
Why is uint64_t needed on a port? Aren't they 8bit?

jmamma:
yes, ports are 8bit on the AVR.

ignore the variable names (copy pasta), it's only used for the 64Bit masks that are used to represent pattern/track steps.

jmamma:
I had a play around with https://godbolt.org/ earlier. It seems whenever we do a shift, an assembly instruction to a subroutine is called, so it's not possible to count the amount of instructions (afaik). 

Probably need to find another way to bench these.

yatli:
shifting a long would require the `__ashldiX` family, but 8bit shifting can be done with `LSL` and `LSR` instructions.

```c
void FN_SET_BIT64(uint64_t& target, uint8_t bit)
{
    uint8_t* pbit = reinterpret_cast<uint8_t*>(&target)+(bit >>3);
    uint8_t mask = bit & 0x7;
    *pbit |= mask;
}
```
```AVR
// let me annotate the things a bit
FN_SET_BIT64(unsigned long long&, unsigned char):
        // [r25, r24] is "&target", 16bits
[...]

PR #25, opened 2019-05-28
Bitarith

yatli:
address #24.


avr/cores/megacommand/CommonTools/helpers.c
```diff
@@ -26,6 +26,9 @@
  * @{
  **/
 
+const uint8_t _bvmasks[]  PROGMEM = {0b00000001, 0b00000010, 0b00000100, 0b00001000, 0b00010000, 0b00100000, 0b01000000, 0b10000000};
```
jmamma: Because we are optimizing for speed, I think these should be loaded in to SRAM.

"I think you will find it takes 3 clock cycles to read from PROGMEM compare to two for RAM"
https://forum.arduino.cc/index.php?topic=134782.0

jmamma: 16bytes vs 1 cycle :thinking:

avr/cores/megacommand/CommonTools/helpers.h
```diff
@@ -32,38 +34,39 @@ extern "C" {
 #define _BV32(i) (uint32_t)((uint32_t)(1UL << i))
 
 /** Bit-level access and test macros. **/
-#define SET_BIT(port, bit)   ((port) |= (1 << (bit)))
-#define CLEAR_BIT(port, bit) ((port) &= ~(1 << (bit)))
-#define TOGGLE_BIT(port, bit) ((port) ^= (1 << (bit)))
-#define IS_BIT_SET(port, bit) (((port) & (1 << (bit))) ? 1 : 0)
-#define IS_BIT_CLEAR(port, bit) (((port) & (1 << (bit))) == 0 ? 1 : 0)
+#define SET_BIT(target, bit)                (((uint8_t*)(&target))[(uint8_t)bit / 8U] |=  _bvmasks[(uint8_t)bit & '\07'])
```
jmamma: Presumably the compiler will optimize
```(uint8_t)bit / 8U```
to
```(uint8_t) bit >> 3```

Confirm?

Comment (yatli) [reply avr/cores/megacommand/CommonTools/helpers.h]: Actually no! This is most surprising to me.

Comment (yatli) [reply avr/cores/megacommand/CommonTools/helpers.h]: See the other thread.

File: avr/cores/megacommand/CommonTools/helpers.c (hunk 1 above)
Comment (yatli) [reply avr/cores/megacommand/CommonTools/helpers.c]: we'd probably remove `PROGMEM` here. The code should not work because I was not using pgm_read_byte (the `LPM` instruction), but a simple `LD` instruction which only works for RAM.

Comment (yatli) [reply avr/cores/megacommand/CommonTools/helpers.c]: you're right, `LPM` is 3cyc.
https://www.microchip.com/webdoc/avrassembler/avrassembler.wb_LPM.html

**Comment by yatli** [comment]
With masks in RAM: `Global variables use 29745 bytes (52%)`

With masks in PGM: `Global variables use 29729 bytes (52%)`

Didn't even scratch it. It's a MEGA anyway ;)

**Review by jmamma — CHANGES_REQUESTED** [review]
uint8_t bit / 8U  is not necessary for 8bit ?

```(((uint8_t*) (&target)) [(uint8_t)bit /8U]```
should simplify to ```(uint8_t ) target```

**Comment by jmamma** [comment]
This review is causing MCL to lock up shortly after boot.

Looking at it now.

**Comment by jmamma** [comment]
Can I push commits to this PR?

I accidentally created a new branch with the fix. 

yatli-bitarith

Cherry-pick if you have to

**Comment by jmamma** [comment]
MCL is booting now but looks like IS_BIT_SET64 is not working, seeing random trigs in the SeqStep page.

**Comment by yatli** [comment]
I think you can push to my branch. I've ticked the option "let maintainers push"

**Comment by jmamma** [comment]
! [remote rejected] yatli-bitarith -> yatli-bitarith (permission denied)

Getting late in the day for me. If you can cherry-pick my commit d81a5e656a7a2e71609ae32042231d615bd4a2d6

**Comment by yatli** [comment]
Sure. Take your time, I'll deal with it when I'm off work

**Comment by jmamma** [comment]
Very weird behaviour. The sequencer is working perfectly which means IS_BIT_SET64 is working fine in ISR.

However, in SeqPage.cpp draw_pattern_mask()
```
291         } else if (IS_BIT_SET64(pattern_mask, i + offset)) {
```

This line appears to be generating random values for some trigs...

**Comment by yatli** [comment]
maybe it is from other parts?
BIT_SET64, BIT_CLEAR64, BIT_TOGGLE64 do not return the full 64-bit value anymore.

**Comment by jmamma** [comment]
I think it's fixed now.

There were compiler warnings 

```/Applications/Arduino.app/Contents/Java/hardware/valence/avr/cores/megacommand/MCL/SeqPage.cpp: In member function 'void SeqPage::draw_pattern_mask(uint8_t, uint8_t, bool)':
/Applications/Arduino.app/Contents/Java/hardware/valence/avr/cores/megacommand/MCL/SeqPage.cpp:287:49: warning: suggest parentheses around '+' in operand of '&' [-Wparentheses]
         } else if (IS_BIT_SET64(pattern_mask, i + offset)) {
                                                 ^
```

yatli:
I've cherry-picked your patch.

jmamma:
Added another patch f901992c11951f21e7164159d46e3be67ba3efe8

yatli:
I've probably seen something like `return BIT_TOGGLE64(....)` in the code?

jmamma:
> return BIT_TOGGLE64

BitArray.hh seems to. 

We can remove that class. it is not used.

yatli:
https://github.com/jmamma/MIDICtrl20_MegaCommand/blob/f402bc211072ebcfc485aa9b74592d91e7a1771d/avr/cores/megacommand/CommonTools/BitArray.hh#L76

It's the 8-bit version. I was paranoid :D

yatli:
note, the 64-bit versions are not compatible with the `return TOGGLE_BIT64(...)` style due to the use of `do{...}while(false)` pattern. It should be safe.

I can further make 16/32bit versions like that. Proceed?

jmamma:
> I can further make 16/32bit versions like that. Proceed?

I forgot what the do / while  guard is supposed to achieve. do you have a c++ reference for me to read?

yatli:
it's simple.
you cannot do:
```
return do { ... toggle bit } while(false);
```

jmamma:
False positive with the fix for the random trig behaviour on line 291. 

It's still happening after I reset SeqPge.cpp

Not sure what changes I made that fixed it. I had a while loop in that function printing the bvmasks, for any memory changes.

yatli:
On a mobile phone now.

The brackets look suspicious: bit / 8 is not isolated

jmamma:
cool. having a jam on MD. will be back tomorrow

yatli:
Have fun!
I’ll test it when I’m back home

yatli:
observation.

In the sequencer, the steps cannot be properly set.
Some steps keep changing. Probably the address is miscalculated and thus we peek into another address.

yatli:
observation.

In the sequencer page:

Step 1 functions normally.
Step 2 does not respond to Trigger 2.
Step 2 responds to Trigger 10. That's one byte away.
Step 3 does not respond to Trigger 3.

However, during playback, it appears all the triggers are correctly set.
So the problem is the sequencer page representation, the `IS_BIT_SET` family.

yatli:
work in progress.

yatli:
Looks like it's working now.
However I notice a problem that the note interface events don't always translate to a trigger toggle.
Tapping a trigger and release sometimes will be ignored by the sequencer -- I have to press the trigger and hold for ~200ms (?) to register.

jmamma:
In MDExploit.h ->
```12 #define TRIG_HOLD_TIME 200```

This is used on the SeqStepPage.

```
245         if ((slowclock - note_interface.note_hold) < TRIG_HOLD_TIME) {
246           CLEAR_BIT64(mcl_seq.md_tracks[last_md_track].pattern_mask, step);
247           mcl_seq.md_tracks[last_md_track].conditional[step] = 0;
248           mcl_seq.md_tracks[last_md_track].timing[step] = 12; // upper
249         }
```

If you like, have a play with different values to see what feels normal to you.

Trig presses on the step page need to
1) Set Trig
2) Clear Trig
3) Select Trig (for manipulation of conditional and microtiming settings)

---

I just noticed on the SeqParam page I have hard-coded this value 😕 

```
160         if ((slowclock - note_interface.note_hold) < 300) {
161           CLEAR_BIT64(mcl_seq.md_tracks[last_md_track].lock_mask, step);
162         }
```

Should create a new PR to make this consistent across all sequencer pages.

**Comment by yatli** [comment]
Sure, we can fix that in another PR.
And I think this is ready for a merge then.

**Comment by jmamma** [comment]
Great. Will test tonight and then merge.

**Comment by yatli** [comment]
Per line 245-249, I think it's for clearing a step that's already set.
However, what I was seeing is that setting a step also requires a press-and-hold, which is not reflected on:
https://github.com/jmamma/MIDICtrl20_MegaCommand/blob/babc8b15f00c22dcd08792d2cc839762e987c445/avr/cores/megacommand/MCL/SeqStepPage.cpp#L242

**Comment by yatli** [comment]
as the trigger push action is perfectly displayed in the seq page, it means note_interface is properly passing the information to the page.

~~so my guess is that SeqPage intercepted the event somehow before it reached SeqStepPage.~~

edit: checked the code. SeqPage will only intercept the event when a button is pressed. So that's not the case.

**Comment by jmamma** [comment]
> so my guess is that SeqPage intercepted the event somehow before it reached SeqStepPage.
Hmm I dont think that's happening 

Looking at that code more closely. Steps are only set/cleared when the trig is released.

I will do some more testing tonight to see If I can replicate what you're seeing.

**Comment by jmamma** [comment]
> However I notice a problem that the note interface events don't always translate to a trigger toggle.
Tapping a trigger and release sometimes will be ignored by the sequencer 

@yatli How infrequent is this?

The code above has a small problem 
```if ((slowclock - note_interface.note_hold) < 300) {```

slowclock is uint16_t. if it overflows it resets to 0.

slocklock = ms. 65,535 / 1000 = 65.35

Every 65 seconds setting a trig might fail.

To avoid the overflow problem I use (when I'm writing code and paying attention 😆 ): 

```
363 /**
364  * Return the difference between old_clock and new_clock, taking into
365  * account overflow of the clock counter.
366  **/
367 uint16_t clock_diff(uint16_t old_clock, uint16_t new_clock) {
368   if (new_clock >= old_clock)
369     return new_clock - old_clock;
370   else
371     return new_clock + (65536 - old_clock);
372 }
```

yatli:
The behavior is more easily triggered on the first steps of the first track.

yatli:
It’s not that infrequent. Triggers almost every time.

jmamma:
I'm not seeing this behaviour.

Wonder if your step 1 trigger is misfiring.

yatli:
hmmmm. I cannot reproduce the problem today.

jmamma:
09e89bde33da27e09784615baf0f9ddbc9eac68c

Uploaded a patch to fix overflow.

Let's leave this one, if you see the problem again we'll create a new issue.

PR #26, opened 2019-05-30
Low Level Optimisations

jmamma:
Going to submit low level optimisations here @yatli 

Anything to speed up the ISR routines.

First one is:

There are only two memory banks, so we don't care what the absolute
value is.


avr/cores/megacommand/MD/MD.cpp
```diff
@@ -162,10 +162,10 @@ void MDClass::triggerTrack(uint8_t track, uint8_t velocity) {
...
 void MDClass::setTrackParam(uint8_t track, uint8_t param, uint8_t value) {
-  if (global.baseChannel > 15)
-    return;
-  if ((track > 15) || (param > 33))
+/*  if (global.baseChannel > 15)
     return;
+    if ((track > 15) || (param > 33))
+    return;*/
 
   uint8_t channel = track >> 2;
```
yatli, line 170: track / 4? now that we know the trick ;)

jmamma: Did I miss something important. I remember you said that 
n / 4 didn't optimize to n >> 2

If you can see optimisation improvements for setTrackParam please assist :smiley:

yatli:
I propose to split RingBuffer into two classes, one for built-in buffer and one for external ptr.

jmamma:
> I propose to split RingBuffer into two classes, one for built-in buffer and one for external ptr.

Yeah that's probably a better approach

avr/cores/megacommand/MD/MD.cpp
yatli: It's not "n/4 didn't optimize to n>>2", rather the opposite:
n/4 will perform better than n>>2 (!)

yatli: >improvements for setTrackParam

sure, will go through that

jmamma: @yatli I worked out what's happening

Both n /4 and n >> 2 will be optimized to the same instruction set when the
-Os compiler flag is enabled

https://godbolt.org/z/sRIPhS

jmamma: we are already compiling with -Os

jmamma: This one is good read. https://ww1.microchip.com/downloads/en/AppNotes/doc8453.pdf

yatli: check this out: https://godbolt.org/z/luMJFr

yatli: When the shifted value is used as addressing index (which must be 16bit), shift generates two more instruction than the division version.

(I've no idea why this is happening... But let's just sacrifice the young goat, for fundamentally technical reasons. :D)

jmamma: 😲

yatli: I just took another look:

```
1:      asr r31
        ror r30
        dec r18
        brne 1b
```

so it's not just two more instructions. it's a loop

😂

jmamma:
i've merged this branch in to dev

PR #31, opened 2019-06-14
RAMPage

jmamma:
This is loosley based on Wesen's breakdown page.

The RAMPage automates sampling the MD, either in mono or stereo.

Tracks 15 + 16 are alternated  between  RAMRecord/RAMPlayback machines.

Top left button -> Record
Top rIght button -> Play/Slice normal
Shift2 -> Play/Slice special.

Encoder 4 -> Record Length (in bars)
Encoder 3 -> Number of slices
Encoder 2 -> Slice mode effect
Slice mode features a bunch of different ways to cut up the sample on playback.
Such as reverse, random re-order, re-order reverse etc... 

--

Im using chain mode to schedule the record and playback tracks. But this may change.

I initially had timing issues switching between RAMR1 and RAMP1 on the same track, bad feedback caused by parameter changes before sample record finished.


yatli:
This is very interesting. I do have a similar use case -- it would be great if track 15/16 can be set up as Inp machines, activated, and switched back to midi channels (I read your tips!) so that I can forward two audio channels from another box, and still make use of the sequencer.

jmamma:
> This is very interesting. I do have a similar use case -- it would be great if track 15/16 can be set up as Inp machines, activated, and switched back to midi channels (I read your tips!) so that I can forward two audio channels from another box, and still make use of the sequencer.

Yeah track stacking!

I think we need to solve the MIDI machine problem more generally. 

Right now cant use MIDI machines with MCL because they override the trigger exploit by transmitting on their own channel. We can probably detect and work around this now, as the current Kit is regularly saved and received by the firmware. 

Also, somehow MIDI machines should be merged in to the internal Sequencer and the output of these machines transmitted on port 2.

jmamma:
@yatli 

May be of some use to you. 

I've added the ability to reflect bitmaps in horizontal and vertical axis. 

Useful for animations. The RAMPage now features spinning tape reel.

jmamma:
fyi: https://javl.github.io/image2cpp/

jmamma:
@yatli 

Have a play with these RAM Pages when you get a chance.
Sample the MD internally, or record a mono synth through InputA or B then slice up the melody.

So much fun!

yatli:
forgot to check in `  uint8_t ram_page_mode;`?

yatli:
forgot to check in `MenuPage ram_config_page(&rampage1_menu_layout, &config_param1, &config_param6);
`?

yatli:
with these two modifications in place I can properly load it up.
whoa!!

one is so much fun and we've got TWO?

minor: is there a way to stop playback? 🤔

yatli:
5AM, cannot stop

jmamma:
There's a bug in MenuPage.cpp causing the dest_var of first menu element to toggle when entering the menu page.

For example, enter RAMPage and Link: setting changes automagically.
Same thing happens in MIDI menu in some circumstances. 

I think it has to do with recycling encoders amongst menus and encoder->hasChanged.

PR #32, opened 2019-06-17
Add support for the Minicommand hardware

jmamma:
In order to achieve this, we must:

MCL Changes:

1) Modularise the MegaCommand code such that we can turn off features to save space.

Core changes:

1) Change timer code/ports to be compatible with ATMEGA64
2) Adjust UART ports for Minicommand (megacommand port names are offest by 1)
    Minicommand does not have a second transmit port so we can disable this.
3) Specify SD Card pins for Minicommand
4) Specify LED pins for Minicommand
5) Re-enable boot_loader interrupt/sysex listener for firmware recieve via midi.
6) Adjust stack memory requirements:

The ATMEGA64 processor only has 4KB of onboard SRAM where as the MegaCommand has 8KB.

We can't use a stack size of 8KB with the ATMEGA64 as the upper 4KB will be unavailable when switching memory banks,


jmamma:
~Timer code looks like it is compatible across micro-processes.~

I've specified the SD Card pin and correct LED config

jmamma:
Timer code needs to be looked at. Timers are not identical.

---

I think we're around 80% complete on getting the core to compile for the minicommand.

Adjusting the MCL firmware will be another story.

jmamma:
http://ww1.microchip.com/downloads/en/DeviceDoc/atmel-2490-8-bit-avr-microcontroller-atmega64-l_datasheet.pdf

jmamma:
This compiles and links.

However, when using the Makefile, the linker appears to be incorrectly pulling in MCL libraries even when "include MCL.h" is removed.

This was a problem I noted in master branch, but has not been a concern until now.

Compiling master with the Arduino IDE does not link in unused libraries.
Thought this might have to do with --gc-section(s) option but no luck.

jmamma:
Makefile is working correctly now.

With WAVDesigner and OLED display disabled we're only 20KB over the 64KB limit.

That's way better than I was anticipating.

jmamma:
I think the Minicommand build will need to remain it's own branch.

Compiling using the Arduino IDE is probably not possible because of conflicts with platform.txt
and the need for different compile options for the different hardware types

jmamma:
``` /bin/avr-size --format=avr --mcu=atmega64 main.elf
AVR Memory Usage
----------------
Device: atmega64

Program:  126572 bytes (193.1% Full)
(.text + .data + .bootloader)

Data:      21822 bytes (532.8% Full)
(.data + .bss + .noinit)
```

^ After disabling the ext tracks,  the firmware compiles and links
However it looks like the firmware size is 126,572 bytes. which is disappointing.

jmamma:
I've tested these changes to make sure that they are compatible with the dev branch and MegaCommmand hardware.

I've now merged these changes in to dev.

Dev branch is now backwards with Minicommand. I have not tested on the Minicommand yet.

For compiling for Minicommand:

Need to disable OLED_DISPLAY and MEGACOMMAND macro definitions.

and use the Minicommand Makefile: minicommand.mk.

PR #33, opened 2019-08-08
Copy/paste revamp, adds clipboard func.

jmamma:
Copy/paste multiple rows using selection rectangle

Encoder 3 = width
Encoder 4 = height

Copied data is written to a temporary file on the
SDCard. 

This means slot data can be copied between
projects.


jmamma:
Testing reveals this works quite well!

PR #35, opened 2019-10-12
SD Drive, refactoring, and various GUI improvements

yatli:
Trying to implement "poor man's +Drive"

PR #47, opened 2019-10-17
GridPage, SeqPage improvements

yatli:
Requires #31 
Requires #35 

Tentative changes to various GUI parts.

GridPage:
- General refactoring
- Adds a visual cue for active slots.
- Use vertical dash line to draw 4-separators.
- [x] Quick edit current slot
  - [x] Quick chain write current slot to workspace
  - ~~[ ] Quick save workspace to current slot (?)~~
  - (Hold, new ideas coming in, need further design)

SeqPage:
- Comments
- SHIFT2+ENC4 change track
- Display active track
- [x] Unify rendering
  - ~~[ ] Expose sequencer cells to derived classes~~
  - ~~[ ] Expose param knobs to derived classes~~
  - [x] Chromatic mode
  - [x] UI fine tuning
  - The current architecture suffices the needs.
- ~~[ ] Display source grid slot that the workspace comes from~~
- ~~[ ] If coming from a source grid slot and modified, show a dirty flag~~
- [x] Info panes

GridIOPage:
- ~~[ ] Suggestion: rename chain to "LOAD", in correspondence to "SAVE"~~
- ~~[ ] When chain write to dirty workspace, keep an old copy for UNDO~~
- (Move to another PR)


yatli:
Conflicts in GridTask and GridPage.
Looks like range-based apply is in place! I'll take a look again and update.

yatli:
conflicts resolved.
- [x] need to resolve: https://github.com/jmamma/MIDICtrl20_MegaCommand/pull/35#issuecomment-543591470

yatli:
testing. some glitches after the merge:

~~on startup initialization (creating config etc.), after the splash screen, the text input page for creating new project does not display correctly~~
~~active slots do not blink~~
~~play status not shown~~
~~midi clock not reflected on led~~
~~transport display not shown correctly~~
~~seqpage track keeps jumping around~~

edit: just shrodinbug coming from the MD side. see commends below.

jmamma:
Can you verify that your build directory is clean.
remove all stale .o files.

jmamma:
Also I had to upload another patch because DEBUGMODE was still enabled. i noticed some weird behaviour, that might be the same as what you're seeing .

jmamma:
Hmm where is the dupe in dev?

```
grep -r ram_config_page .
./MCLMenus.cpp:MenuPage ram_config_page(&rampage1_menu_layout, &config_param1,
Binary file ./MCLMenus.o matches
./MCLMenus.h:extern MenuPage ram_config_page;
./MCLMenus.h:        {"RAM Page" ,0, 0, 0, (uint8_t *) NULL, (Page*) &ram_config_page, (void*)NULL, {}},```

**Comment by yatli** [comment]
nope, I added the dupe myself yesterday for testing :)

**Comment by jmamma** [comment]
okay. it's late here. brain no work no more.

**Review by yatli — COMMENTED** [review]
Went over the code.
No obvious gotchas.
I then did a factory reset on MD and uploaded MCL, and all is well again.

**Comment by yatli** [comment]
The new grid slot menu is wonderful.
What do you think if we allow range select even in non-menu mode?

**Comment by yatli** [comment]
When you have time, please try this branch -- I've got some of the workflow ideas baked in.
Specifically, you can now directly edit the slot under cursor with SHIFT2+ENC Push.

**Comment by jmamma** [comment]
Will do. im thinking of moving sequencer pages in to quick select.

**Comment by yatli** [comment]
> moving sequencer pages in to quick select

Any idea what to do with the freed encoder buttons?

**Comment by jmamma** [comment]
Nope 😄. But it might help with standardising the GUI.

Encoder button should be fast scroll.

**Comment by yatli** [comment]
> fast scroll

+++

**Comment by yatli** [comment]
tracking: #56

**Comment by jmamma** [comment]
I had the same idea(s) for the sequencer.
Instead of round buttons, square with circle LED

I like the idea of having the side panel with information displayed.  
I also think a full screen mode is good too.

One thing to note is that manipulating any 'on screen' MD parameter(either via MD encoder or via MIDI cc) will cause the trig interface to break after around 5-10seconds.  This is true with the exception of the track master LEV.

**Comment by jmamma** [comment]
Some more thoughts. If we move sequencer pages to QuickSelect.
shift1 must always be reserved for quick_select page. 

Need to assign another button for exit, Save

That only leaves two buttons Shift2 and Write.

**Comment by yatli** [comment]
Some annotation:

Sequencer_large

**Comment by yatli** [comment]
I think the sequencer pages can be consolidated into this single page.

Note page = NOT
Lock page = LCK
RTRK page = "Record sign"

chromatic page is not a sequencer page.

How about:

WRITE = RTRK
SHIFT2 = NOTE/LOCK
SHIFT2+WRITE = Turn knobs for LOCK target changes

The rationale being, once set up, probably we don't need to change lock targets as frequently.

**Comment by jmamma** [comment]
You know that chromatic page has realtime record capabilities ?
Those are tied to the sequencer.

**Comment by yatli** [comment]
No. (⊙_⊙;)

**Comment by jmamma** [comment]
Yeah that was the whole point.

Save button enables record mode. 

You can overdub. It's pretty cool way to make melodic patterns on the fly.

**Comment by jmamma** [comment]
Also you can record legato on the A4. Something the A4 sequencer doesnt support.

**Comment by yatli** [comment]
How about four minor modes.
NOTE | LOCK | LOCK TARGET | CHROMATIC

Then 
WRITE = RTRK
SHIFT2 = Minor mode switch
SHIFT2 + Knob press = Quick minor mode switch

(
  SHIFT2 + Knob1 rotate = MD/Ext/A4
  - or -
  SHIFT2 + WRITE = MD/Ext/A4
)

This aligns well with current SHIFT2+TRIGS/KNOB4 = Track select, and Knob press = quick minor mode switch

**Comment by yatli** [comment]
> One thing to note is that manipulating any 'on screen' MD parameter(either via MD encoder or via MIDI cc) will cause the trig interface to break after around 5-10seconds

I've noticed that, the Seq page will exit sometimes. Can we detect that, and automatically recover?

**Comment by jmamma** [comment]
So currently there is a CC hook, to exit if CC are sent from MD -> MC when TriggerExploit is active.

```
312 void MDExploitMidiEvents::onControlChangeCallback_Midi(uint8_t *msg) {
313   if ((msg[1] < 120) && (msg[1] > 15)) {
314       if (md_exploit.state) {
315         route_page.update_globals();
316       }
317       if (md_exploit.off()) {
318         GUI.setPage(&grid_page);
319     }
320   }
321 }
```

I thought about recovery. 

You dont want to re-engage the exploit each time a CC is received as this will just flood the MD with  sysex.

Might be able to re-engage exploit every X seconds after CC received.

There is a delay of a few seconds between when parameter is changed and the exploit fails. I havent measured this time, so I dont know if it is consistent.

**Comment by yatli** [comment]
> Might be able to re-engage exploit every X seconds after CC received.

maybe everytime CC is received, activate a blocking infobox about recovery, that latches a few hundreds ms?

```
CC message
infobox shown, timer set to 0ms
...
CC message
timer reset to 0ms
...
timer timeout
recover
```

**Comment by yatli** [comment]
> There is a delay of a few seconds between when parameter is changed and the exploit fails

A random guess: when a param is changed on screen, there's an animation and a latch timer. Could be correlated?

**Comment by jmamma** [comment]
Do you want to investigate and see if you can find out something I may have missed?
🕵

**Comment by yatli** [comment]
I'd like to first work on the SeqPage GUI revamp. Will keep an eye on this issue.
Is that ok chief? :)

**Comment by jmamma** [comment]
ha no problem. Plenty to do. I'm juggling Encoders, LFOs and some MidiClock stuff

**Comment by yatli** [comment]
> How about four minor modes.
> NOTE | LOCK | LOCK TARGET | CHROMATIC
> ...

Ok that doesn't work. Here's a list of mandatory interactions:

- Switch seq pages (which I like to call 'minor modes', whereas 'major modes' are larger concepts, like seq, ram, etc.)
- Quickselect menu (major mode)
- Exit to grid
- Real-time tracking ON/OFF
- MD/Ext select
- Sequencer page selection

A possible assignment:

```
SAVE = Exit
SHIFT1 = Major/minor mode selection, a 2-level menu
SHIFT1 + Knob press = minor mode quick selection (seq pages, ram machine id etc.)
WRITE = RTRK
SHIFT2 = Sequencer page selection / track selection
SHIFT2 + Knob press = 4 functions, one being MD/Ext toggle
```

@jmamma what do you think?

**Comment by jmamma** [comment]
I think it's a good start. I like the idea of shift2 + encoder for sub page selection. For example.
Minor modes make absolute sense for sections like WavDesigner. In which it consists of 4 sub pages. 

From your description above it sounds like Sequencer Section is selected using Major Mode and then the sub pages are chosen using Minor modes?

I'm a little hesitant in hiding pages behind a 2nd layer of 'selection' unless absolutely necessary.
Specifically I feel that any page that provides real/time performance capabilities (like editing a step sequence) should be no more than 2 button presses away.

I will think more about this tonight.

**Comment by yatli** [comment]
merged FX pages for encoders

**Comment by yatli** [comment]
@jmamma when you have time.

Looks pretty cool :)

**Comment by jmamma** [comment]
Wow looking really good so far. Amazing work.

I have one gripe. The circle sequencer buttons. The MD != OCTATRACK. 
That is all :-).

**Comment by jmamma** [comment]
Another small suggestion. On the GridPage, the first dotted vertical line needs to be shifter to the left 1 pixel. This means we might have to get rid of the row indicator

I had trouble with this area of the GUI. if elements are touching the GUI feels cramped.

**Comment by jmamma** [comment]
One more thought we might want to consider for GridPage.

I had an idea to have the first or last Row of the Grid to always show the tracks that are currently loaded. 

This could be made to disappear and the full grid exposed when rotating the vertical encoder.

This has interesting implication because it means that slots could be made to load tracks from other columns.

Thoughts, or too complicated?

**Comment by jmamma** [comment]
FXPage PR was merged in to dev, so pull those changes in here for working encoder gfx and improved TomThumb numerics

**Comment by yatli** [comment]
tick, tick, 🚀

**Comment by yatli** [comment]
> This has interesting implication because it means that slots could be made to load tracks from other columns.

I'm trying to understand this (why does this mean that).
Sounds like copying a slot to another (obviously I didn't get it)

**Comment by jmamma** [comment]
> Sounds like copying a slot to another (obviously I didn't get it)

That would be one way to achieve the same thing. Copy + Paste during a live performance is not ideal.  There is a performance scenario currently that you might not want to overwrite tracks in slots 1-4, but the tracks you want to load are in those columns. I used to have a method of holding shift to do an offset load. I am considering adding it back. 

There is also the possibility of making Chain mode intelligently load slots to "free tracks" on the MD.

Also, In chain mode you could instruct slots to automatically load tracks from other columns in the grid.

In addition. RAMPage has functionality that it loads special slots not accessible to the user. This 'current slot' view could be a way of illustrating this.

**Comment by jmamma** [comment]
MD mute menu uses number under every 4th trig, to identify the trigger positions more clearly.
I dont think we have space for numbers, but a marker at every 4th trig might be useful in improving clarity.

File: avr/cores/megacommand/MCL/SeqPage.h [hunk 1]
```diff
@@ -27,20 +27,68 @@ class SeqPage : public LightPage {
...
+  void select_track(uint8_t device, uint8_t track);
+
+  virtual bool handleEvent(gui_event_t *event);
+  virtual void loop();
+  virtual void display();
+  virtual void setup();
+  virtual void init();
+  virtual void cleanup();
+
+  static constexpr uint8_t pane_x1 = 0;
```
jmamma, line 56: These constant variables are going to be stored in RAM?

Might be better to change them to macro defined constants.

yatli: These are const expressions and therefore inlined whenever used.
Think about them as the C++ answer to macros.

yatli:
Extended the knob frames downward 1 pixel as the beat marker.

yatli:
The chromatic page feels more natural to me because of the global menu. It switches to the recording circle on RPTC.

nit: the display update flag does not respond to play/stop state, so it'd stuck in "play" or "stop".

jmamma:
LEN missing from Chrom page?

Also: I'd like to draw a chromatic keyboard below the encoders.

(I can't push to your branch)
```
mcl_gui.draw_keyboard(32, 23, 6, 8, 24, 0);

void MCLGUI::draw_keyboard(uint8_t x, uint8_t y, uint8_t note_width, uint8_t note_height, uint8_t num_of_notes, uint64_t note_mask) {
  uint16_t chromatic = 0b0000010101001010;
  uint8_t note_type = 0;

  for (uint8_t n = 0; n < num_of_notes; n++) {
     if (IS_BIT_SET16(chromatic, note_type)) {
     oled_display.fillRect(x - 1,y + 1, 3, (note_height / 2) - 1, WHITE);
     }   
     else {
     oled_display.drawRect(x,y, note_width, note_height,WHITE);
     x += note_width - 1;
     }   
     note_type++;
     if (note_type == 12) { note_type = 0; }
  }

}
```

note_mask will be used to draw active notes. ^

jmamma:
Regarding the display_update on record pages. This was disabled to improve record latency.
Theres probably a better way to negate the latency. Specifically tag note_on messages with a MidiClock timestamp.

yatli:
@jmamma I added deferred rendering for updating the keyboard highlight. It compensates the latency issue.

yatli:
@jmamma I've implemented the ext seq page. This will need your help to test -- I don't have A4 hooked up at the moment, and haven't figured out how to input notes.

I suggest we feature-freeze this PR and move the things (Grid current row etc.) to another one.

jmamma:
> This will need your help to test

Will do.
--

Yes, I agree with feature-freeze. I'd like to merge this one in to dev, but only when all the original sequencer functionality is accessible.

Is the SeqRlck page still missing?

jmamma:
ExtSeqPage looks good!

I had originally envisioned having 4 rows of rectangles displayed, one per note to illustrate polyphony. This can be added at another time. 

Thanks you.

jmamma:
Can you cherry-pick this one ~e370bfbbf8689f17cc105d4a87c5153058b9cadf~

973bc0e45151bd9fff2fe0de0da18b2746989998

jmamma:
Not sure why but "Sharp/Black" notes from MIDI2 are not being highlighted in draw_keyboard()

jmamma:
MD External control mode, is not using the octave + scale data to remap MIDI2 note_num.
Updated code should look like: 
But implementation is still buggy.

```
void SeqPtcMidiEvents::onNoteOnCallback_Midi2(uint8_t *msg) 
...
519   if ((mcl_cfg.uart2_ctrl_mode - 1 == channel) ||
520       (mcl_cfg.uart2_ctrl_mode == MIDI_OMNI_MODE)) {
521     uint8_t pitch = seq_ptc_page.calc_pitch(note_num);
522     seq_ptc_page.trig_md_fromext(pitch);
523     SeqPage::midi_device = midi_active_peering.get_device(UART1_PORT);
524     return;
```

Same patch needs to be applied for onNoteOffCallback_Midi2

yatli:
973bc0e is in.

yatli:
>SeqRlck

Working on it :)

yatli:
TODO:

in ext seq page, there are 8 pages instead of 4. Should update the base implementation to add a configuration variable.

yatli:
~~@jmamma `calc_pitch` or `seq_ext_pitch`?~~

edit: figured out. it's in `trig_md_fromext`

yatli:
~~I should have fixed the external control logic.~~
edit: hooked up A4, not working -- on A4 keyboard input the firmware just reboots.

yatli:
>   uint8_t scaled_pitch = pitch - (pitch / 24) * 24;

what does 24 mean?

jmamma:
12 notes in an octave. With NUM_KEYS ~36~ 32 we can fit 2 full octaves of notes.
2 x 12 = 24.

jmamma:
Basically in order for the "pressed" key notes to be in range of draw_keyboard() you need to remove the octave offset (otherwise they wont be drawn)

For example we should translate. C3->C0 and C2->C1.

Does that make sense?

yatli:
Yes. I was thinking about changing that 24 to NUM_KEYS but it didn't work. It should be aligned to octaves.

jmamma:
> Yes. I was thinking about changing that 24 to NUM_KEYS but it didn't work. It should be aligned to octaves.

yeah I made the same error 😮

yatli:
I've connected A4 today. Didn't realize I can use the A4 keyboard before :)

yatli:
Is it possible to use A4 as clock source?
My MD has some clock drifting issues.

jmamma:
What kind of clock drifting issues?

MIDI -> CLK REC: MIDI2 will get you going (though I think there is a problem with clock 2 as source, tempo calc  fluctuates periodically.)

yatli:
My MD's clock keep getting faster and faster when running. Perhaps the crystal is aging, or the temperature compensation has failed..

yatli:
MIDI -> CLK REC: MIDI2 doesn't forward the clock to MIDI1. There's no such option in CLK SEND

jmamma:
A few suggestions regarding the sequencer.

When a parameter lock step is reached by the sequencer, the "LED" above the trig does not flash fully. Single pixel instead of fullRect. Looks weird to me.

I'm wondering if we should make the lock LEDs blink. similar to the MD when a lock is made. 

I feel that a MD step should flash when it is played (if it has a trig on it). ***

We can use [x] to represent some other feature of the sequencer. I'm thinking about adding slides in at at a later date.

jmamma:
> MIDI -> CLK REC: MIDI2 doesn't forward the clock to MIDI1. There's no such option in CLK SEND

This is on the MC.

System -> MIDI -> CLK REC

----

All MD MIDI settings are controlled by the MC.

jmamma:
> My MD's clock keep getting faster and faster when running.

Does this happen when MIDI is unplugged?

Could be your right arrow button getting stuck? 

🤷‍♂

jmamma:
RTRK page should have length control (like RLCK)

jmamma:
note_mask should be cleared when scale/octave params are changed. probably worth clearing it on init() too.

yatli:
> Could be your right arrow button getting stuck?

Nak, it doesn't change that fast.
I had reported this issue because I thought it was a MCL issue, but it turns out it's just my MD.

jmamma:
> I had reported this issue because I thought it was a MCL issue, but it turns out it's just my MD.

what model. what firmware you running?

yatli:
> what model. what firmware you running?

It's a SPS-1 UW MKII without +Drive.
Running firmware 1.63.

jmamma:
Maybe try a soft reset. backup your work first.

yatli:
Thanks for the suggestion. I've tried to reset a couple of times.
The behavior is interesting, the firmware still thinks the tempo is stable at the set value, but the actual clock has drifted. So if I boot up at 130bpm, connect MC, the tempo got synced over as 130.
After a while, MD still displays 130 but on MC it measures the real tempo and the value gradually goes bigger and bigger.

That's why I think it's not a firmware issue but the physical crystal.

jmamma:
Hmm weird.

I have noticed.
If you send tempo change via SYSEX, the display doesn't update immediately.

yatli:
> but the physical crystal

if so it means my MD is slightly "overclocked" and thus more powerful. hah.

jmamma:
> My MD's clock keep getting faster and faster when running. Perhaps the crystal is aging, or the temperature compensation has failed..

~Clock can't go out of sync with itself. Maybe some corrupt RAM?
Or issue with interprocessor communication/timing~

I think I misunderstood you here. Just to clarify, does your A4 show the same drift when slaved directly to the MD. (no MC involved)

jmamma:
It's late here. Tomorrow i'll merge this review.

I think next step is sorting out GUI button assignment. 

Then push to master for a release.

I'm trying to think of ways to make firmware upload easier. A simple tool to upload the .hex + multiplatform. I was looking for a WebApp solution  but havent found anything that works.

yatli:
okay, have a nice one.

yatli:
I'm still trying to figure out why A4 keyboard input crashes the firmware.

jmamma:
> I'm still trying to figure out why A4 keyboard input crashes the firmware.

uint64_t note_num ? array out of bounds?

SET_BIT/CLEAR_BIT?

yatli:
ok got it, it's the channel setting -- added guards for that.

yatli:
fixed the rough edges.
if everything works as expected, this PR would be ready.
see you tomorrow :)

yatli:
> I'm trying to think of ways to make firmware upload easier.

A web-based solution may not work, because it doesn't have access to the USB peripherals or serial ports.
A cross-platform native executable could be the way to go.

That means Qt, Electron, GTK, Avalonia etc.

yatli:
next time I'll put the PR branch in this repo.

jmamma:
> fixed the rough edges.
> if everything works as expected, this PR would be ready.
> see you tomorrow :)

Awesome work @yatli .

Your help here is greatly appreciated. The new GUI is fantastic and matches my original vision 👀 

> A web-based solution may not work, because it doesn't have access to the USB peripherals or serial ports.

Chrome WebApps had access to serial, but it looks like they're deprecated now.

PR #51, opened 2019-10-19
FX Pages.

jmamma:


jmamma:
@yatli I have added MCL_GUI::draw_encoder(..)

Draw's an animated MD style style encoder. 
Need to create different display modes.

MD track style encoder.
MD global fx page style encoder
A4 style encoder.

etc.

yatli:
I think it's possible to just store one copy of each type of encoder image and then manually draw the dial marker

jmamma:
Yeah I thought about that. I wasn't going to write the algorithm for the snake like dial marker.

Thought about using draw @ radius but overhead using sin/cos not worth it.
I also looked at the drawCircle algorithm for calculating dial x,y position. which is interesting. 

After all that I went with sprites 😃

yatli:
sorry it's just my PGM space phobia 😅
It's just a little bit over 100 bytes anyway. We can totally afford this.

yatli:
I should have killed ~30KB PGM space yesterday.
tick, tock, and we're now in the "tock" phase 😁

jmamma:
We will have PGM space for the next 100 years after yesterday 🚀

jmamma:
also. another good thing, is that get_random_byte no longer retrieves 0s  😆

jmamma:
Hmm. Performance of the FX Pages actually seems very good.
I was worried about conflict with the sequencer. But with the ISR blocking to copy SYSEX into MIDI TX buffer, it's all okay.

This makes me think we might be able to pull off LFOs for FX pages.
Side chain compression, without using FX tracks.

jmamma:
OxF0              | 1
MD HEADER  | 5
DATA              | 3
0xF7               | 1

10 bytes per FX message

192 callbacks. = 12 * 4 = 48 interrupts per beat.

124 bpm = 2bps
0.5 seconds per beat @ 124bpm.

2 * 48 * 10 = 960 bytes per second.  

8x Bandwidth = (31250 * 8) / 10 = 25KB/s      

(10bits per byte need to be transmitted via USART)

jmamma:
LFO proof of concept works. 

Simple RAMP LFO. Adjusting delay level on beat.
```
void MCLSeq::seq() {

  //  if (in_sysex == 0) {

  //  for (uint8_t i = 0; i < 1; i++) {
  //    lfos[i].seq();
  //  }

  for (uint8_t i = 0; i < num_md_tracks; i++) {
    md_tracks[i].seq();
  }

  //  }
  // if (in_sysex2 == 0) {
#ifdef EXT_TRACKS
  for (uint8_t i = 0; i < num_ext_tracks; i++) {
    ext_tracks[i].seq();
  }
#endif

  if (MidiClock.step_counter == 1) {
    my_lfo = 48;
  }
  MD.sendFXParam(MD_ECHO_LEV, my_lfo, MD_FX_ECHO);
  if (my_lfo < 126) {
    my_lfo += 2;
  }
}
```

jmamma:
Need to create parameter names for FX machines and in turn use those for master fx names.

yatli:
forgot to check in `uint16_t LightPage::encoders_used_clock[GUI_NUM_ENCODERS];`?

jmamma:
@yatli This is working now.

The light encoders work and behave exactly like the A4 encoders/MD FX encoders.

~Theres a small bug where the encoder Value is re-displayed when the slowclock repeats for the nth  time.~ <- fixed

jmamma:
Also I noticed Tom Thumb numerical chars dont match MD/A4. I will alter this.

PR #59, opened 2019-10-25
LFOs

jmamma:


jmamma:
Getting my head around how these will be implemented in to workflow.

I like the idea of having Global LFOs for Side Chain behaviour. 
It would be nice to have multiple destinations either FX or track param.

LFOPage needs to show LFO waveform.
Have Trig interface for step sequence if LFO is TRIG type.
Have destination values for the LFO.
Have Speed control. Fastest speed is 1/128 note resolution (same as MD LFOs)
Waveform repeats every 10.667 steps.

If LFOs are global then they wont be loadable via chain mode.

An alternate would be Global LFOs, + track LFOs. All available track LFO waveforms would need to be in RAM for fast playback. RAM usage = number_of_wav * 127.

jmamma:
experimenting with LFO_LENGTH=96 samples.
There are 96 sequencer ISR events per 2 beats.

jmamma:
All the functionality for LFOs is now implemented.

LFO Page gives access to a single LFO with 2 selectable destinations.
Destination can either be MDTracks [1-16] or MD FX Tracks [17-21]
Destination 0 = off.

Current MD parameter values are retained at clock start and on CC change, much like Parameter Lock pages. This means the LFO value is relative to the position of the MD parameter.

LFO behaviour varies according to Waveform. Symmetrical Waveforms (SIN + TRI) oscillate around the original parameter value whilst, exponential waveforms use the original parameter values as the MAX.

LFOs can be either free running, reset on trig with loop playblack. or reset on trig with single cycle playback. Each LFO has a dedicated sequencer track for entering in reset trigs.

Currently there is a single LFO depth for both destinations. This is beacuse I wanted to avoid multiplication scaling in the sequencer interrupt. Instead the WavTable is recalculated when depth is changed.

jmamma:
Realised we can leverage the RingBuffer positions to accurately determine transmission rates. 

Added a diagnostic page to show throughput speed on Midi Ports.
Currently showing Byte/s and Buffer Size  for TX and RX buffers.

Will add nice "speedometer" style graphics in the future.

jmamma:
Note: rx buffer will not be accurate as sysex bypasses the ringbuffer.

yatli:
If we can afford one cycle in the interrupt, we can host a uint8_t counter in there. It doesn't have to keep exact bytes transmitted, as we want very coarse metric.

PR #62, opened 2019-10-29
Midi clock

jmamma:
By counting to 4 instead of 6, 

div128_time = time_for_4_pulses / 8 

which can be calculated by shifting instead of rcall and eliminate floating point operation.

Will probably save 1000 cycles.

### DIFF
--- a/avr/cores/megacommand/Midi/MidiClock.h
+++ b/avr/cores/megacommand/Midi/MidiClock.h
@@ -47,9 +47,9 @@ public:
   volatile uint8_t mod3_counter;
 
   volatile uint8_t mod8_free_counter;
-  volatile uint16_t div128_counter;
-  volatile uint16_t div128_time;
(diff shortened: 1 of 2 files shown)


jmamma:
I've lost my faith in ⌛️

yatli:
What's happening? 😯

jmamma:
None of the gear in my studio produces stable clock 😝

jmamma:
We can actually code in MC internal clock if you think this will be useful. 
START/STOP Transport will need to be received from external device, but clock source can be generated internally. 

It's a few lines of code. The functionality is already there.

yatli:
I'd still prefer an external clock -- don't want to have a "tempo page" :D

yatli:
> None of the gear in my studio produces stable clock 😝

`T      i   c         k.` 😅

PR #66, opened 2019-11-04
add comments on MidiClock

yatli:
@jmamma I still haven't figured out how the 5KHz timer and MIDI_CLOCK both affects div192th and mod12counter.

### DIFF
--- a/avr/cores/megacommand/Midi/MidiClock.cpp
+++ b/avr/cores/megacommand/Midi/MidiClock.cpp
@@ -32,7 +32,6 @@ void MidiClockClass::init() {
   clock_last_time = clock;
   mod12_counter = 0;
   mod6_counter = inmod6_counter = 0;
-  mod3_counter = 0;
   bar_counter = 1;
   beat_counter = 1;
   step_counter = 1;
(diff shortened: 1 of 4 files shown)


jmamma:
div192 is a intermediate clock event generated at t / 2 , where t = time between midi time pulse.

To get an 'average' time. We calculate the time taken for 8 MIDI pulses then divide by ( 2 * 8 = 16 ) to get the interpolation time.

```
div192_time = diff_clock8 / 16;
```

div96th counter is increased on MIDI clock message. 
div192th counter is increased on MIDI Clock message and when (MidiClock.div196th_countdown >= MidiClock.div196_time);

We always want to use MidiClock signal where we can. Half of the div192 events occur on MidiClock, the other half we use the interpolation time.

```
X=div96
x=div192

X     X     X     X
x  x  x  x  x  x  x
```

jmamma:
Eek. just realised I renamed div192 to div196.

Must have been half asleep.

yatli:
As my MD clock is going crazy I'm now receiving clock from A4.
Is it possible to still use MD to send transport messages (play, stop)?

jmamma:
> As my MD clock is going crazy I'm now receiving clock from A4.
> Is it possible to still use MD to send transport messages (play, stop)?

This feature is not implemented. It would require additional logic in MidiUart.cpp .
Currently transport is linked to same port as clock.

jmamma (commented):
I'd like to merge this.

PR #70, opened 2019-11-09
PageSelectPage wip

yatli:


jmamma:
@yatli the last commit implements what I had in mind for pageselect changes.

Split the 16 triggers in to 4 groups

Trigs [0 - 3]  
  -  0: Grid
  -  1: Mixer
  -  2: Route
  -  3: LFO

Trigs [4 - 7 ]
   - 4: Step Edit
   - 5: Record
   -  6: Params
   - 7: Chromatic

Trigs [8 - 13]
  - 8 Sound Browser
   - 9: Wav Designer
   -  10: Loudness

Trigs [14 - 15]
  - 13: DELAY
  - 14: REVERB
  - 15: RAM1
  - 16: RAM2
(RAM tracks are mapped to the actual tracks that are manipulated when recording/playing from the respective RAM Pages.

--

I've started disabling Encoder[1-4] buttons for exiting to GridPage.
Shift 2 now becomes page select in all relevant pages.

Care needs to be taken when switching pages whether to disable then re-enable the trigger interface. 

I need to work out the details here. It might not be worth disabling the Trig interface if it's to be re-enabled immediately on the next page load. If the trig interface is to be disabled then the Global Settings may need to be retransmitted to make sure the Audio routing is consistent across the two globals if a routing change has occurred. Some juggling needs to be performed here as transmitting globals can cause brief sequencer dropout.

yatli:
I've implemented quick select in category. Use encoder buttons to jump within the cat.
The category name is now displayed and fits 1602 LCD.

jmamma:
> I've implemented quick select in category. Use encoder buttons to jump within the cat.
> The category name is now displayed and fits 1602 LCD.

Cool i'll check this out later tonight.

--

I was thinking that Button 1 should setPage(&grid_page)

jmamma:
Probably should draw_trigs() on PageSelect page to give visual indication that that the trig interfaces can be used.

jmamma:
@yatli 

I think we should change encoder[0] range to category size.
And encoder[1] range to number of categories. 

Can we change encoder behaviour from relative to absolute. I don't see the need here unless I missed something.

yatli:
> Can we change encoder behaviour from relative to absolute. I don't see the need here unless I missed something.

The pages are not numbered consecutively.

yatli:
Also, for category, the cat id is first determined with the page index of the current page (we only know the page number, not index), so that's also relative.

yatli:
@jmamma can you revert the display code? It was intended for the legacy displays.
I'll be working on a new visual layout tonight for OLED.

jmamma:
> Also, for category, the cat id is first determined with the page index of the current page (we only know the page number, not index), so that's also relative.

Right. 

Can we still achieve ->

```
I think we should change encoder[0] range to category size.
And encoder[1] range to number of categories.
```

jmamma:
The original HD44780 display code is still there.
Just change the OLED code.

```
184 void PageSelectPage::display() {
185 #ifdef OLED_DISPLAY
186   oled_display.clearDisplay();
187   char str[16];
188   oled_display.setCursor(0,0);
189   oled_display.print("Page Select:");
190   get_category_name(page_select, str);
191   oled_display.print(str);
192 
193   oled_display.setCursor(0,15);
194   get_page(page_select, str);
195   oled_display.print(str);
196   oled_display.display();
197 
198 #else
199   GUI.setLine(GUI.LINE1);
200   char str[16];
201   GUI.put_string_at_fill(0, "Page Select:");
202   get_category_name(page_select, str);
203   GUI.put_string_at(12, str);
204 
205   GUI.setLine(GUI.LINE2);
206   get_page(page_select, str);
207   GUI.put_string_at_fill(0, str);
208 #endif
209 }
210 
```

yatli:
Oh okay, overlooked that one. :)

yatli:
@jmamma the concept art:

image

jmamma:
Cool!

I was debating whether or not Popup should be used.

Need to see if it looks too similar to Save/Chain pages.

yatli:
If using a popup, there's no space for an icon.
Thoughts?

jmamma:
> If using a popup, there's no space for an icon.
> Thoughts?

Ah yeah.

Perhaps we go for blank page layout, and try and put the icons in.

yatli:
Second try:

image

Better horizontal space utilization. More futuristic. :)

jmamma:
beauty.

yatli:
I'll cleanup and implement this.

yatli:
@jmamma it's done.

jmamma:
Cool. ill test now.

We need to decide new button layout for FIle Browser and Menu.

Eliminate the use of Encoder buttons would be ideal.

Save: NO/BACK. Write YES/FORWARD ?

jmamma:
@yatli looks really good. nice job

yatli:
> We need to decide new button layout for FIle Browser and Menu.
> Eliminate the use of Encoder buttons would be ideal.

Yeah. Let's try and unify the key assignments.
The functionalities (and current key):
- Select entry (Enc push)
- Menu (S2)
- Page select (N/A, but should be "return to machinedrum config" in SDDrive, and "return to GRID" on config menu)
- Back / Up one level (N/A)

Also we should consider:
- system menu keys
  - Select entry (Enc push)
  - Menu (S2)
  - Back / Up one level (SAVE and S1)
  - Exit (N/A)
- wait_for_confirm popup menu button layout
  - YES (S2)
  - NO (S1)
- Text input
  - Clean (WRITE)
  - Case (S2)
  - InputPane (S1)
  - Cancel (SAVE)
  - Confirm (Enc push)

jmamma:
Select entry (Enc push) -> I think we should eliminate encoder button completely where possible.

yatli:
We probably need some kind of table to help designing the layout.
Excel?

jmamma:
I was hoping there was some online tool for this exact purpose.

jmamma:
8011a07

implements my above suggestion.

There is a complication in how you access the SYSTEM MENU, and the Save / Chain pages.

Save/Write pages are opened on BUTTON_RELEASED

System Menu is exited on BUTTON_PRESSED.

When you Exit System menu you immediately trigger a BUTTON RELEASE in the GridPage and open the SAVE  page.

yatli:
Let's try this:

https://1drv.ms/x/s!Aql6jSQH9uyZhPpDVrfyjYOCSa56sw

jmamma:
@yatli We're going to need a way of changing tracks from the MC sequencer pages.

Because the Trig interface is active in the PageSelect page it won't be possible to switch tracks unless exiting back to Grid.

I'm thinking [ Shift 2] opens up a track menu with 

- Track Selection, 
- Clear Track
- Clear All Tracks
- Track Resolution/Scale, 
- Step Shift, 
- Copy Track, 
- Paste Track,

Thoughts?

yatli:
Yeah that's what I thought about.
One comment is about Clear current track.
 If it's moved from WRITE to the menu, it'll be a bit harder to invoke, which slows down real time dubbing/clearing workflow.

A workaround is to have the menu "sticky" -- if one selects clear track, the selection is not reset to the first entry when re-entering the menu.

S2 can then be thought of as a reconfigurable key -- by default it does what the user set last time. If last action is clear track, then it stays clear track, until reconfigured with enc2

yatli:
Current progress:

image

S1 has a very consistent press-and-hold sense.
S2 is mostly ex-menu, or backspace in text edit (could also be a menu, include backspace/clear/case).

The grid page S2+Enc push was "load and edit with seq page", currently commented out.
Possible to add an option to re-enable this?

jmamma:
Added:
wavd_icon.png
loudness_icon.png

improved:
lfo_icon.png
mixer_icon.png (matches display of mixer page)

jmamma:
Nice work @yatli !

jmamma:
```
169   if (EVENT_RELEASED(event, Buttons.BUTTON3)) {
170     encoders[0] = &seq_param1;
171     encoders[1] = &seq_param2;
```

Because we do this, the underlying sequencer page draw_knob() values change as encoders 1 and 2 are rotated when show_seq_menu == true.

If I recall, my solution for GridPage was to hardcode the encoder pointers in the sequencer pages member functions.

Unless there is a better way?

jmamma:
Track Select, offset by 1 so that the number are 1->16.

Also, track select beyond 5 doesn't appear to work

jmamma:
- [x] Combine Clear Trk and Clear all In to one menu item with 2 options "Trk" and "All"

jmamma:
Page Select Icon's are so beautiful 😂

yatli:
@jmamma I'm burning out. It's still a bit messy, could you help to clean it up a bit?

jmamma:
Will look at it tomorrow. Need to eat dinner and have a few hours of rest. 

Thanks for your effort!

yatli:
Incorrect behaviors:

- [x] Track select offset +1
- [x] Resolution changing not working
- [x] Dynamically configure track select enc range for ext
- [x] On RTRK/RLCK, the dynamic menu config has no effect. They should have different entries enabled.
- [x] On PARAMS/CHROMA, menu scrolling is a bit weird -- selected entry always on the first row?
- [x] On STEP, len does not persist

yatli:
When operating the menu, it's easy to trigger S2 + ENC3, which is currently change all track length.
I'm changing it to WRITE + ENC3.

yatli:
Cleaning up code, powered by zen inertia. I've no idea what I'm writing but it fixes problems!

yatli:
It's taking the shape now. I'll leave the stubs (copy, paste etc.) to you.

yatli:
ALL DONE

yatli:
I've compiled the current key assignment table:
image

It's in the second sheet in the excel link.

jmamma:
@yatli 

I'm wondering if SoundBrowser should be moved from PageSelect in to TrackMenu.
Kind of makes more sense as it is related to the current track.

Thoughts?

jmamma:
TrackMenu vs GridSlotMenu

Different behaviour relating to clear / copy / paste

In grid slot Menu. 
```
Clear: [ -- , YES ]
Copy: [ -- , YES ]
Paste: [ -- , YES ]

When you close the menu options revert back to --
```

Where as Track Menu
```
Clear: [ TRK , ALL ]
Copy: [ TRK, ALL ]
Paste: [ TRK, ALL ]
```
The side effect of this is that when you open up the track menu, it may already be positioned over Clear, and releasing the button may clear your track or all your tracks unintentionally.

Should we use the GridSlotMenu method? Or should we use the TrackMenu method, but add a confirmation dialog box? 

The other option is to reset menu item position to "Track Select", but this feels weird and slows down workflow imo.

yatli:
> Should we use the GridSlotMenu method? Or should we use the TrackMenu method, but add a confirmation dialog box?

I prefer GridSlotMenu method.
As an extra safety guard, we can implement undo in another PR.

> I'm wondering if SoundBrowser should be moved from PageSelect in to TrackMenu.

Hmm. Considering the workflow, when editing the patch MC automatically exits to the GRID.
If the sound browser is activated via TrackMenu, then it's going to either pull an untouched MD patch, or, go from GRID -> Seq -> TrackMenu

Put it in the GridMenu maybe?

jmamma:
@yatli copy/paste is working for MDTracks*

I decided to re-use MCLClipboard. Means the copied tracks are not stored in RAM but on the SDCard.

give it a test when you have some time.

jmamma:
Left Right shifting is working, but parameter locks are not being copied correctly.

Anything obvious here:

```

519 void MDSeqTrack::rotate_right() {
520 
521   ROTATE_RIGHT(lock_mask, length);
522   ROTATE_RIGHT(pattern_mask, length);
523   ROTATE_RIGHT(oneshot_mask, length);
524 
525   int8_t new_pos = 0;
526 
527   MDSeqTrackData temp_data;
528 
529   memcpy(&temp_data, this, sizeof(MDSeqTrackData));
530 
531   for (uint8_t n = 0; n < length; n++) {
532      if (n == length - 1) { new_pos = 0; }
533      else { new_pos = n + 1; }
534      memcpy(&locks[0][new_pos], &(temp_data.locks[0][n]), NUM_MD_LOCKS);
535      conditional[new_pos] = temp_data.conditional[n];
536      timing[new_pos] = temp_data.timing[n];
537   }
538 
539 }
             
```
@yatli 
?

jmamma:
ah memcpy won't work. because of the array order.

jmamma:
@yatli SeqPtcPage.

info1 should display A4->T[1-4] or MI->T[1-4] when it's in Ext mode.

midi_device is detected/changed when TI is detected from MD.

458     // note interface presses are treated as musical notes here
459     if (mask == EVENT_BUTTON_PRESSED) {
460 
461       SET_BIT64(note_mask, pitch);
462       midi_device = device;

Or when notes are received on UART2 

570   SeqPage::midi_device = midi_active_peering.get_device(UART2_PORT);

Need to update labels on midi_device change, whilst not adding additional latency.

jmamma:
@yatli 
```
 29 const menu_t<7> seq_menu_layout PROGMEM = {
 30     "SEQ",
 31     {
 32         {"TRACK SEL:", 1, 17, 0, (uint8_t *)&opt_trackid, (Page *)NULL, opt_trackid_handler, {}},
 33         {"COPY:", 0, 3, 3, (uint8_t *)&opt_copy, (Page *)NULL, opt_copy_track_handler, { {0, "--",}, {1, "TRK."}, {2, "ALL"}}},
 34         {"CLEAR:", 0, 3, 3, (uint8_t *)&opt_clear, (Page *)NULL, opt_clear_track_handler, { {0, "--",}, {1, "TRK."}, {2, "ALL"}}},
 35         {"CLEAR:", 0, 3, 3, (uint8_t *)&opt_clear, (Page *)NULL, opt_clear_locks_handler, { {0, "--",}, {1, "LCKS."}, {2, "ALL"}}},

```
How does the overloading of CLEAR work here?
Some kind of magic?

yatli:
I've added menu dynamic config when working on the file browser.

Check `SeqPage::config_as_trackedit` for usage.

jmamma:
I added a "Step Menu" which is accessible by holding trig and pressing Button3

Which has two items, COPY STEP, PASTE STEP.

Untested.

jmamma:
We hit 200KB 🎂

jmamma:
Does it make sense to have a Copy Locks routine for the Lock Page. Or keep it as Copy Tracks.

yatli:
> Does it make sense to have a Copy Locks routine for the Lock Page. Or keep it as Copy Tracks.

Makes sense. I guess people won't notice the difference until they observe that notes are not copied sometimes.

yatli:
Briefly tested.

- Draw keyboard on ptc select: stuck and keep being displayed on some tracks.
- Track menu cannot be recalled when the lock menu is triggered

yatli:
keyboard drawing + Enc4 is very intuitive. Love it.

yatli:
> Draw keyboard on ptc select: stuck and keep being displayed on some tracks.

With trigger down I can see the PTC is bouncing between "--" and "E2"
Some kind of stackoverflow / memory leak?

yatli:
The keyboard should be cleared immediately (and then the notes + locks displayed) when the trigger is released.

jmamma:
> The keyboard should be cleared immediately (and then the notes + locks displayed) when the trigger is released.

yes! i haven't gotten around to testing yet. have been emulating the code in my head. 😄

jmamma:
@yatli ^

jmamma:
Previously, when in GridPage. before entering sequencer pages and before calling page_select.
grid_page.prepare() was called.

```
587 void GridPage::prepare() {
588   if (MD.connected) {
589     MD.getCurrentTrack(CALLBACK_TIMEOUT);
590     MD.currentKit = MD.getCurrentKit(CALLBACK_TIMEOUT);
591     if ((mcl_cfg.auto_save == 1)) {
592       MD.saveCurrentKit(MD.currentKit);
```

which was responsible for getting the latest kit state before entering sub pages.

Adding prepare to page_select.init() adds lag. 

wondering when best time to call this would be.



Need to think of best time to call this routine.

jmamma:
I think I could keep patching this forever.

yatli:
wow that's a lot of incoming changes.
will test again tonight.

yatli:
```
C:\Users\Yatao\Documents\Arduino\hardware\MIDICtrl20_MegaCommand\avr\cores\megacommand\MCL\LFOPage.cpp: In member function 'virtual void LFOPage::display()':
C:\Users\Yatao\Documents\Arduino\hardware\MIDICtrl20_MegaCommand\avr\cores\megacommand\MCL\LFOPage.cpp:275:51: error: no matching function for call to 'LFOPage::draw_page_index(bool, int)'
```

forgot to checkin the signature?

jmamma:
Merging this, as core page_select functionality has been implemented.

yatli:
🙌

PR #102, opened 2020-01-20
Implement TrigInterface class

jmamma:
@yatli 

Compiles, haven't uploaded yet.

I've only made the change to PageSelect page. The other pages will still attempt to use md_exploit.

### DIFF
--- a/avr/cores/megacommand/MCL/MCL.cpp
+++ b/avr/cores/megacommand/MCL/MCL.cpp
@@ -43,8 +43,9 @@ void MCL::setup() {
 
   mcl_actions.setup();
   mcl_seq.setup();
-  MDSysexListener.setup(&Midi);
 
+  MDSysexListener.setup(&Midi);
+  trig_interface.setup(&Midi);
 #ifdef EXT_TRACKS
   A4SysexListener.setup(&Midi2);
 #endif
--- a/avr/cores/megacommand/MCL/MCL.cpp
+++ b/avr/cores/megacommand/MCL/MCL.cpp
@@ -47,7 +47,6 @@ void MCL::setup() {
   MDSysexListener.setup(&Midi);
   trig_interface.setup(&Midi);
   md_track_select.setup(&Midi);
-  md_track_select.on();
 #ifdef EXT_TRACKS
   A4SysexListener.setup(&Midi2);
 #endif
--- a/avr/cores/megacommand/MCL/MCL.cpp
+++ b/avr/cores/megacommand/MCL/MCL.cpp
@@ -37,7 +37,7 @@ void MCL::setup() {
   MidiClock.setTempo(mcl_cfg.tempo);
 
   note_interface.setup();
-  md_exploit.setup();
+  //md_exploit.setup();
 
   MD.midi_events.enable_live_kit_update();
 
--- a/avr/cores/megacommand/MCL/MCL.cpp
+++ b/avr/cores/megacommand/MCL/MCL.cpp
@@ -47,6 +47,7 @@ void MCL::setup() {
   MDSysexListener.setup(&Midi);
   trig_interface.setup(&Midi);
   md_track_select.setup(&Midi);
+  md_track_select.on();
 #ifdef EXT_TRACKS
   A4SysexListener.setup(&Midi2);
 #endif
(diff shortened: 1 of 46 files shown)


avr/cores/megacommand/MCL/MCL.cpp
```diff
@@ -43,8 +43,9 @@ void MCL::setup() {
 
   mcl_actions.setup();
   mcl_seq.setup();
-  MDSysexListener.setup(&Midi);
 
+  MDSysexListener.setup(&Midi);
+  trig_interface.setup(&Midi);
```
yatli, line 48: Suggest:

Use 0x30 to query if TI mode is present.
Then multiplex md_exploit within trig interface?

jmamma: Yeah that might be a good idea.

Would be even better if users upgrade their machines  :)

yatli:
~~uploaded and not working.~~

trying 
eb97899

jmamma:
It is sort of working. 

First trig press works. Subsequent actions dont update the trig number, I think the old sysex message is being "replayed"

yatli:
The first press switches to the wrong page.
Misaligned payload?

jmamma:
70   uint8_t trig = sysex->getByte(2);

trying.

jmamma:
fixed.

yatli:
can confirm, fixed.

yatli:
ridiculous low latency.

jmamma:
> ridiculous low latency.

Yeah, end_immediate() is why,

jmamma:
So now that this is working. Tomorrow I will port the rest of the MCL code to use TI interface instead of md_exploit, with a possible fallback mechanism.

I think we need to have a few different TI modes.

Is it possible to hide the sequencer event LED flashes ?

yatli:
wrt trig leds:

overlay (XOR), or override?

yatli:
yeah, exactly this question.
I'll implement override now

jmamma:
Override needed for step edit.

XOR for mixer page?

--

Do we get the MC to tell the MD what LEDs to light up, or maybe in some modes the MD can do that for us.

For example, step edit, we need to send the steps across (bitmask maybe?) 

But for mixer page let the MD work out which LEDs should be set.

jmamma:
It's late here. i'm sure theres a simpler implementation to what i'm asking

jmamma:
sleep time.

yatli:
good night!
tmr I'll be catch a plane. It's Chinese Spring Festival here, off for a few days with no access to the gears.

yatli:
> Do we get the MC to tell the MD what LEDs to light up, or maybe in some modes the MD can do that for us.

the logic can be implemented in MegaCommandTI directly.

jmamma:
> good night!
> tmr I'll be catch a plane. It's Chinese Spring Festival here, off for a few days with no access to the gears.

No worries, enjoy your break.

avr/cores/megacommand/MCL/MCL.cpp
jmamma: I've decided to deprecate md_exploit.

Having both systems will add complexity to the code in which I do not want to manage. 

For example, there are some scenarios whereby the exploit must be turned off, during (specific sysex commands) and then turned back on again.

jmamma: Also, it means we have to test both systems during fw upgrades. No thanks.

yatli: I agree. Let's move forward.

yatli: It also means that exploit-related bugs are all gone -- when you tweak the params, the TI no longer fails; no need to switch to track 16 anymore;

jmamma:
```jmammarella@DESKTOP-DCRTMLQ:/mnt/c/Users/Justin/dev/mdhack/fw$ cat loader.s
        .text
        .align  2

        .globl  MegaCommandLoader

MegaCommandLoader:
  pea 0x002c0000
  pea 0x000e96e0    /* base image size = 0xe56e0; +offset 0x4000 */
  jsr 0x000001c4    /* UNPACK_FIRMWARE from e96e0 to 2c0000 */
  addq.l #0x8, %sp
  jsr 0x002c0000    /* MegaCommandInit (fw.text base address) */
  jmp 0x00213a0c    /* Resume MD_MAIN */
```

Mystery solved, MegaCommandInit address is hardcoded.

PR #104, opened 2020-02-26
Sample management

yatli:
Work in progress.

- .spk sample packs in SD-Drive
  - [x] Push
- .syx in SoundBrowserPage
  - [ ] Send (NOT WORKING)
- SampleManager with gateway 0x34
  - [x] List, implementation in FileBrowserPage
  - [ ] Rename, erase, copy/paste
- Better .wav file support
  - [x] extended attribute
  - [x] multi-channel wav files
  - [ ] smpl chunk
    - [x] loop points
    - ~~[ ] root node~~
    - [ ] slices
- ~~WavEdit~~


yatli:
@jmamma couldn't get the sysex listener to work.

Here's my attempt at sample slot querying:

```c++
void SoundBrowserPage::query_sample_slots() {
  DEBUG_PRINT_FN();
  encoders[1]->cur = 0;
  encoders[1]->old = 0;
  numEntries = 0;
  cur_file = 0;
  uint8_t data[3] = {0x70, 0x34, 0x00};
  sysex->addSysexListener(this);
  for (int i = 0; i < 48; ++i) {
    mcl_gui.draw_progress("Loading slots", i, 47);
    s_query_returned = false;
    MD.sendRequest(data, 3);
    while (!s_query_returned) {
      handleIncomingMidi();
    }
    add_entry(s_samplename);
  }
  sysex->removeSysexListener(this);
}
```

... and `end_immediate` is never reached.

I initialize it like this:
```c++
  SoundBrowserPage(Encoder *e1 = NULL, Encoder *e2 = NULL, Encoder *e3 = NULL,
                  Encoder *e4 = NULL) : FileBrowserPage(e1, e2, e3, e4), MidiSysexListenerClass() {
    ids[0] = 0;
    ids[1] = 0x20;
    ids[2] = 0x3c;
  }
```

and expect to have a pattern matching in `end_immediate`:

```c++
void SoundBrowserPage::end_immediate() {
  DEBUG_PRINT_FN();

  if (sysex->getByte(0) != 0x00)
    return;
  if (sysex->getByte(1) != 0x20)
    return;
  if (sysex->getByte(2) != 0x3C)
    return;
  if (sysex->getByte(3) != 0x02)
    return;
  if (sysex->getByte(4) != 0x00)
    return;
  if (sysex->getByte(5) != 0x72)
    return;
  if (sysex->getByte(6) != 0x34)
    return;
  DEBUG_PRINTLN("SoundBrowserPage: end_immediate passed");
  for (int i = 0; i < 5; ++i) {
    s_samplename[i] = sysex->getByte(7 + i);
  }
  s_samplename[5] = 0;
  s_query_returned = true;
}
```

What am I doing wrong?

jmamma:
``` 7 class TrigInterface : public MidiSysexListenerClass {
  8 
  9 public:
 10   bool state = false;
 11 
 12   TrigInterface() : MidiSysexListenerClass() {
 13     ids[0] = 0x7F;
 14     ids[1] = 0x0D;
 15   }
 16   void setup(MidiClass *_midi) {
 17     sysex = &(_midi->midiSysex);    <----- you're missing him
 18   }

 trig_interface.setup(&Midi);
```

Need to set which MidiSysexClass to attach to. 
Either Midi->Sysex or Midi2->Sysex

otherwise:
```
  sysex->addSysexListener(this); 
```
wont work.

Should probably set default for Midi->Sysex, as this has caught me out a number of times.

jmamma:
Also,
```
  if (sysex->getByte(0) != 0x00)
    return;
  if (sysex->getByte(1) != 0x20)
    return;
  if (sysex->getByte(2) != 0x3C)
    return;
```

Are validated from

```
    ids[0] = 0;
    ids[1] = 0x20;
    ids[2] = 0x3c;
```
before calling end/end_immedate.

no need to check them twice.

yatli:
Sending wav from sd to MD is working now.
@jmamma could you help figure out how to receive?
Also, the samplemgr file menu is glitching on init. Turn the selection knob and it'll be cleared.

yatli:
Loading my sd card with samples now 🎵~

jmamma:
> @jmamma could you help figure out how to receive?

What do you mean exactly?

jmamma:
> Loading my sd card with samples now 🎵~

Keep them under 48khz 😉

yatli:
Currently I have:

image

But not sure how to proceed:

- Add the sds sysex listener?
- Data packets are processed in end(), not end_immediate(), will it trigger if I keep calling handleIncomingMidi?
- Is it ok to refactor sds sysex listener to allow custom receive file?

jmamma:
sendDumpRequest.. we need to implement this on the MD side.

yatli:
huh, it's simply not there?

jmamma:
Yes, not implemented.

I hacked it using macros.

See   MDClass::send_sample();

yatli:
ha. auto hotkey.

jmamma:
Also. MCL is already configured to automatically receive samples.
```
  6 void MCL::setup() {
...
 55   MidiSDSSysexListener.setup(&Midi);

```

See:
```
119 void MidiSDSSysexListenerClass::dump_header() {
```

Currently the filename is set using the sample receive position. But this can be rewritten to use the sample_name[] array which gets populated before SDS dump is received.

yatli:
what's the difference between setup and `addSysexListener`?

jmamma:
> what's the difference between setup and `addSysexListener`?

setup tell the sysexlistener child class which parent sysexlistener to attach to. either Midi->Sysex or Midi2->Sysex.

```
292 void MidiSDSSysexListenerClass::setup(MidiClass *_midi) { sysex = &(_midi->midiSysex); sysex->addSysexListener(this); }

```

Add / remove actually adds the callback to the Listener Parent, so that when sysex dump is received the childs end_immediate() or end() functions are called.


---

To refactor this. Setup and add should probably be combined in to a single function.

yatli:
Samplemgr display is fixed.
Requires https://github.com/yatli/mdhack/commit/ef7d28cd00b10f02a676bd004f68a6b5795a3acc

Give it a go when you have time :)

yatli:
send sysex implemented for SD-Drive.
Need testing.

yatli:
MD stuck at "RECEIVING SAMPLE" popup:

[Bd1.zip](https://github.com/jmamma/MIDICtrl20_MegaCommand/files/4265999/Bd1.zip)

44.1Khz, 16bit, 2ch

Is it because it's 2ch?

yatli:
Also, I tried to send sysex with SD-Drive.
Target file is MD-909 slot 1.
When transmission starts, MD samplemgr stuck at waiting.
After a while it gives an error like "memory full"

yatli:
so currently, all I can send is "render.wav"

yatli:
Update: it does receive. Just very slow. ~5 packets per second.
The `delay(200)` thing?

yatli:
It's been sending a chained sample for 1 hour. Slow, but pretty stable I have to say!

jmamma:
> The `delay(200)` thing?

Can't remember why that is there. It should send using the handshake protocol.
Edit: maybe openloop was failing when i was testing and I added that in for good luck. You can remove and see what happens. But it should default to closed loop.

Do you see "closed" or "open" loop on the MD side?

yatli:
wasn't in samplemgr, only see a huge popup with a progress bar.
I added some debug info -- doesn't look like it's open loop because I don't see "response time out switch to no handshake"

yatli:
After 1hr I've loaded a 1.7MB sample chain.
Having fun :D :D :D

jmamma:
SDS is slow. but not that slow!

What's your turbo speed? I suspect if it was only doing 5 packets a second it was running in open loop and the delay(200) was kicking in. That delay should be removed.

jmamma:
also if you open sample mgr up before hand. and manually select slot for recv, you should see what transfer mode is in progress.

jmamma:
another possibility is debug mode is causing the timing to mess up and it falls back to open loop.

yatli:
Let me try a release build.

yatli:
Stereo sample always results in "too big" error?

jmamma:
oh yeah. SDS is mono only. 

who is throwing the error? MD or MC?

yatli:
MD is throwing error.

yatli:
Okay, channel-wise too big.
Got it. :)

jmamma:
yeah i didnt add support for stereo samples.

either send 1 channel, or sum L + R needs to be implemented

yatli:
Worked it out.
On 1st `waitForMsg`, it may send back `SDS_WAIT`, which is not checked.
The routine then goes directly into open loop (and thus delay(200))

yatli:
Also, need addSysexListener in waitForMsg.

yatli:
Multi-channel sample supported now

jmamma:
Nice work handling the channel interleaving.

avr/cores/megacommand/Midi/MidiSDS.cpp
```diff
@@ -53,6 +53,7 @@ void MidiSDSClass::sendDumpRequest(uint16_t slot) {
 }
 uint8_t MidiSDSClass::waitForMsg(uint16_t timeout) {
 
+  MidiSDSSysexListener.sysex->addSysexListener(&MidiSDSSysexListener);
```
jmamma: I think this is not the right place to add sysex listener.

MidiSDSSysexListenerClass should already add itself as a listener on setup. The above code would result in listener duplication ?

292 void MidiSDSSysexListenerClass::setup(MidiClass *_midi) { sysex = &(_midi->midiSysex); sysex->addSysexListener(this); }

jmamma (changes requested):

avr/cores/megacommand/Midi/MidiSDS.cpp
yatli: let me test a little bit.
I thought the handler wasn't hit earlier.

yatli: works :)

yatli:
Currently `sendWav` doesn't handle file names.
Any ideas how to implement that?

jmamma:
274 void MDClass::setSampleName(uint8_t slot, char *name) {

yatli:
It kind of works...
After SDS dump finishes, MD will be busy "cleaning up".
If we send the set name request now, it may be ignored, or it may become "sticky" and assign the name to the next transmitted sample 0.0

yatli:
Consider extended dump header?

jmamma:
> It kind of works...
> After SDS dump finishes, MD will be busy "cleaning up".
> If we send the set name request now, it may be ignored, or it may become "sticky" and assign the name to the next transmitted sample 0.0

Hmm. I wonder how the C6 program pulls it off ? 
It may suffer from the same issue though...

jmamma:
When the MD sends a sample it sends the Sample name first immediately followed by the SD.

Maybe try reversing the order ?

yatli:
hmmm interesting. Let me have a try.

yatli:
It works. I didn't expect this is how it should!

yatli:
Check out `0x0020e01e` -- that's how it sends sds dump data packets
(version 46)

The reason we didn't find it earlier is that it's only triggered via an RPC, at `0x00231d32`

yatli:
bulk sample transmission feels good!

yatli:
I'm guessing that ROM machines work just like RAM machines, but the content is sourced from flash.
SDS dump data packets are written to flash directly.
"Cleaning" means maintaining the file blocks.
"Loading" means flash->battery backed SRAM.

Think about a MD fully loaded with 48 ROM samples. If it's loaded from FLASH on the fly, it will take ~10 seconds.
Instead, this is probably done at samplemgr in a one-time service manner. The battery-backed SRAM content won't change until next sample transmission?

jmamma:
> I'm guessing that ROM machines work just like RAM machines, but the content is sourced from flash.
> SDS dump data packets are written to flash directly.
> "Cleaning" means maintaining the file blocks.
> "Loading" means flash->battery backed SRAM.
> 
> Think about a MD fully loaded with 48 ROM samples. If it's loaded from FLASH on the fly, it will take ~10 seconds.
> Instead, this is probably done at samplemgr in a one-time service manner. The battery-backed SRAM content won't change until next sample transmission?

Yeah this is my interpretation. For faster loading we could skip the writing to SRAM bit. 


--- 

Also. From the manual:

```All sample speeds from 4 kHz to 48 kHz are accepted and handled by the Machine- drum UW. Higher rates are accepted but the samples are downsampled to half the sampling speed. The Machinedrum UW does not handle stereo samples. All bit depths are accepted but will be played back at 12 bit.```

https://www.elektronauts.com/t/prepping-samples-for-md/36489/29

I was curious as to what happens to 24bit samples, we didn't clearly identify a truncation. But users are reporting they are converted from 24->16bit (33% reduction in size on save).

**Comment by jmamma** [comment]
I just realised that saving as .WAV we lose any loop data in the SDS header.
Which means a SDS restore wont restore single cycle wavs correctly

.WAV has support for CUE points, which we can encode loop data. 

C6 used to support this:
http://www.elektron-users.com/index.php?option=com_fireboard&Itemid=2&func=view&id=26236&catid=9

This might also be useful for encoding playback slice positions.

http://archive.li/UjwDW#selection-2578.2-2578.5

**Comment by yatli** [comment]
Yep I remember some of the wav samples are loop-enabled.
Wrt slice positions: MD only has 128-segment resolution, and I tried some sample chains I made for OT, doesn't align perfectly.
We may need to pad the elements.

**Comment by jmamma** [comment]
This might be useful ?

```
The ROM machines ROM-25 to ROM-48 operates slightly differently since they are optimized for loops. How the parameters differ from the other ROM machines is described below.
HOLD - Set to a value of 127 the amplitude envelope is open for 2 bars. A setting of 1 will keep the envelope open for 1⁄4 of a step (1/16 note)
STRT - This parameter is now linear which makes it easy to find a position within a loop. In a 2-bar loop you will find each 1⁄4 note located every 128/8 = 16 parameter steps. So for every 16 steps of the parameter you will find each beat in the loop.
END - Also linear and functions basically the same way as the STRT-parameter.
```

RAM machines also behave this way ^

**Comment by yatli** [comment]
I thought this is the case for all ROM machines? 😮

**Comment by jmamma** [comment]
Nope, slots < 25 are exponential

STRT - Controls the start point of the sample. The parame- ter is exponential, giving a very fine control over the begin- ning of a sample.
END - Controls the end point of the sample. The sample will be reversed if it is set to a lower value than the STRT parameter

**Comment by yatli** [comment]
> This might also be useful for encoding playback slice positions.

Consider OT metadata format?
There are free editors for that.

**Comment by jmamma** [comment]
I'm not familiar with the OT metadata format. 
I'm assuming it's the same WAV format as mentioned above?

**Comment by yatli** [comment]
When you save metadata on OT (BPM, gain, slices etc.) it will save a .ot file next to the sample. It's a binary format that is well understood by 3rd party devs.

**Comment by jmamma** [comment]
Feels sacrilegious mixing MD and OT 😁 

That reminds me, at some point I was going to create a WAV editor viewer for the 128x32 OLED 😁 

WavDesigner needs a bit of work too (now that sample transfer does not lockup MD)

**Comment by jmamma** [comment]
Also, we need a quick way of moving RAM recording to free ROM slot.

**Comment by yatli** [comment]
Need to update the sample slot query to also indicate whether the slot is empty.
This can be done by calling `samplemgr_find_slot_start_block_offset` -- when it returns 0 it means the slot is free.

**Comment by yatli** [comment]
The MD SDS dump sequence:
[SDS Dump Header]
[UW SetName]
[SDS Dump Data...]
🤔

**Comment by jmamma** [comment]
> 
> 
> The MD SDS dump sequence:
> [SDS Dump Header]
> [UW SetName]
> [SDS Dump Data...]
> 🤔

Is that from MIDI dump?

**Comment by yatli** [comment]
I get this information from the code directly.

**Comment by yatli** [comment]
Project browser is broken.

**Comment by yatli** [comment]
Now requires https://github.com/yatli/mdhack/commit/f09810d08bbe52aa7b5c52e78ebe7f914088df25.

Sample slot query is done in a batch, and is almost immediate now.
Also detects [EMPTY] slots.

But I'm a bit stuck here. On 1st call everything is fine. On second call the control flow breaks for unknown reason, and returned to the caller:

```c++
void SoundBrowserPage::query_sample_slots() {
  DEBUG_PRINT_FN();
  encoders[1]->cur = 0;
  encoders[1]->old = 0;
  numEntries = 0;
  cur_file = 255; // XXX why 255?
  cur_row = 0;
  uint8_t data[2] = {0x70, 0x34};
  call_handle_filemenu = false;

  sysex->addSysexListener(this);
  MD.sendRequest(data, 2);
  DEBUG_PRINTLN("slot query sent");
  auto time_start = read_slowclock();
  auto time_now = time_start;
  do { // <------------ somewhere here, it doesn't go into the while loop, but acts as if "return" is called ?
    handleIncomingMidi();
    time_now = read_slowclock();
  } while(!s_query_returned && clock_diff(time_start, time_now) < 1000);

  if (!s_query_returned) {
    add_entry("ERROR");
  }
  ((MCLEncoder *)encoders[1])->max = numEntries - 1;

  sysex->removeSysexListener(this);
}

```

jmamma:
s_query_returned is not being reset to false ?

jmamma:
Also can we move away from the proprietary .spk archive and instead use directory structure on the SD card to store WAVs? 

Each directory could be a possible sample bank.

This allows managing banks offline when SD card is connected to computer and allows for sending individual samples easily.

yatli:
> s_query_returned is not being reset to false ?

brain failure!
last night was dreaming about avr interrupts damaging the call stack

yatli:
> Also can we move away from the proprietary .spk archive and instead use directory structure on the SD card to store WAVs?

A .spk is just a txt file listing the file names one by one. Wav files are stored in the filesystem separately.

yatli:
@jmamma the sample management send part is ready.

All sample receive functionalities not implemented.
SYSEX send is not working.

yatli:
Debug builds starting to fail.
Have to remove some debug messages...

yatli:
MidiSDS::sendWav now directly read loop info from the wav file, so wavDesigner doesn't pass SDS loop information anymore.
Instead, WavDesigner saves a wav file with loop info baked in a smpl chunk with one loop descriptor.
Tested working.

yatli:
Testing with loops extracted from a Roland SC-88 soundfont now...

yatli:
It works now, the MachineInst

yatli:
There is an alignment issue though -- loop point position off by one?
I hear "zzz" when it's looping.

jmamma:
> There is an alignment issue though -- loop point position off by one?
> I hear "zzz" when it's looping.

There might be no easy fix here. 

In WAV designer **for single cycle waveforms** I get around this by making sure the fundamental frequency of the looped waveform sample rate line up.  The sample rate is adjusted if thats not the case.

yatli:
Good enough for fast-paced arps at least!

yatli:
Tested a few instruments, wurly, rhode, piano, guitar etc.
The alignment problem is sample-wise -- some of them have perfect playback.

jmamma:
can you upload some samples for me to test. will look at this later tonight.
currently preparing for the apocalypse.

Melbourne is on track for epidemic in next 2 weeks.

yatli:
See: https://github.com/yatli/mdhack/issues/40

yatli:
Take care, stay safe.

jmamma:
Switching brain from 'disaster survivalist' back to 'mcl firmware developer'

😓

jmamma:
@yatli wow excellent work with the sample chunk support!

---

I've tested uploading a few of the instrument samples.

Loop point seems to just be the length of the WAV, is this intentional ?

yatli:
Em no?
Try `wurly/wurly_m-d2.wav`?

jmamma:
checking...

yatli:
The SC-8820 folder also works

jmamma:
ha that works great! i actually didnt know the MD could  pull that off.

jmamma:
Some of the samples I sent did default to looping the WAV length. maybe this was their designated loop point?

jmamma:
i think we need to go ahead and create the waveform editor with loop point support. 😁

yatli:
I unpacked the .sf2 with an ancient tool sf2comp: http://www.hammersound.net/mirrors/last_night/sf.htm

Which sample? I can check with a more advanced tool (e.g. Viena).

jmamma:
SC-88\SAMPLE_13.wav

you may need to increase decay length to hear the transient repeat.

yatli:
image

So yeah, "intentional". :)

jmamma:
layering sample instruments with the ARP is glorious!

jmamma:
we need to add "LA synthesis" to the WAVDesigner. Transient segment + loop segment.

yatli:
Root note information is not yet taken into consideration.
The data is there already, and is well-formed in the samples, see `smpchunk_t::dwMIDIUnityNote` (60 = C4)

yatli:
A workaround for the "zzz" is to go lo-fi.
BRR=30, FLTW=80

jmamma:
How do we use the root note information, on the chromatic keyboard display ?

jmamma:
Yeah filtering will remove the buzz, in a musical production it will be less noticeable. often those sounds fade in frequency spectrum of  other instruments like high hats etc.. 

See "Auditory masking"

yatli:
> How do we use the root note information, on the chromatic keyboard display ?

Chroma keyboard, and step edit note pitches. I think currently the pitch values are based on the assumption that PTCH=64 <-> "Middle C" ?

jmamma:
> 
> 
> > How do we use the root note information, on the chromatic keyboard display ?
> 
> Chroma keyboard, and step edit note pitches. I think currently the pitch values are based on the assumption that PTCH=64 <-> "Middle C" ?

yeah that sounds right

jmamma:
MCL file size 216914 KB now.

Something to keep in mind.

yatli:
Yeah. Debug builds are failing.

yatli:
> Chroma keyboard, and step edit note pitches. I think currently the pitch values are based on the assumption that PTCH=64 <-> "Middle C" ?

That brings another question. How do we establish a link between sample slots, and their original wav files on SD drive.

yatli:
I've been thinking about this for a while.
One possible attack vector is that MD only displays uppercase sample slot names.
For example, "BD01" will display correctly, but "Bd01" will cut off, shown as "B   " in the kit editor.

This means that if we further limit the sample name to 3 characters, we gain one byte (7bit though) that can be associated with 128 sample slots in a project.

Then if we query the sample name and get `"BD1"+0x01`, we then look up the slot 0x01 which reveals that root note, the original wav file, etc.

yatli:
Actually we don't need 128 slots.
48 will be enough

yatli:
Also, about chroma poly voices. How to configure it so that I have 4x4-voice rompler?

jmamma:
multi-timberal. good luck :D

jmamma:
im wondering if we drop support for the A4 to gain back bytes...

yatli:
I wonder how much would we gain back if so.
Some 2% or 3%?

yatli:
Some samples are missing from the file browser.
Name too long? Case sensitivity?

jmamma:
not sure. you'll need to dump the raw listing and see.

If there's no more free space for DEBUG disable EXT_TRACKS in memory.h

avr/cores/megacommand/MCL/Wav.cpp
```diff
@@ -270,6 +270,9 @@ bool Wav::read_samples(void *data, uint32_t num_samples, uint32_t sample_index,
     return read_data(data, read_size, position + data_offset);
   }
 
+  if (num_of_channels > 1) {
```
yatli: `if (num_of_channels == header.fmt.numChannels)`, and can be merged with the above `header.fmt.numChannels == 1`

jmamma: hmm. maybe we do.
if  (channel == 255) then read all channels. 

?

yatli: honestly I think we only have two practical cases, mono files and stereo files.

jmamma: what about 5.1 surround ? 😉

yatli: no need for an API to extract channel 1 and 3 from a quadraphonic file.

yatli: > what about 5.1 surround ? wink

Yeah, I'm wondering what's your intention for loading multi-channel file. Wave editing?

jmamma: yes. wav editor. will probably have a stereo mode. need to find peaks in both channels. 
Much faster to read the stream directly than to buffer each channel.

jmamma:
@yatli ^ have fun.

avr/cores/megacommand/MCL/FileBrowserPage.h
```diff
@@ -9,7 +9,7 @@
 #include "SeqPage.h"
 #include "Menu.h"
 #include "MenuPage.h"
-#include "Midisysex.hh"
+#include "MidiSysex.hh"
```
yatli: The devs are caught using case-insensitive file systems.

yatli:
Updated the makefile a bit to detect system avr-gcc

yatli:
Having problems with the linux build.
The built `main.hex` is only 72787 bytes.
Things not getting compiled?

jmamma:
I havent built on linux for a while. but it should "just work"

jmamma:
You're probably not linking MCL.h

check mcl_setup.cpp 

ie:

```
  1 
  2 #include "MCL.h"
  3 
  4 void setup() {
  5   mcl.setup();
  6   return;
  7 }
  8 
  9 void loop() {
 10   // put your main code here, to run repeatedly:
 11 }
~

**Comment by yatli** [comment]
The `setup` and `loop` in `weak_funcs.cpp` prevented it to be linked properly.

**Comment by jmamma** [comment]
hmm we probably dont need them. this is old ruin&wesen code

**Comment by yatli** [comment]
loaded. the wavform looks a bit strange to me. trying to make sense of the page now..

**Comment by jmamma** [comment]
> loaded. the wavform looks a bit strange to me. trying to make sense of the page now..

They should be accurate. I've compared a few with output on Audacity on PC>

I've added a few patches to improve zooming. Waveform was also inverted.

See: https://github.com/jmamma/MIDICtrl20_MegaCommand/pull/111

**Comment by yatli** [comment]
After merging `dev`:

Release: 209950 bytes (82%)
Debug: 229818 bytes (90%)

**Comment by yatli** [comment]
@jmamma branch rebased to latest dev.

**Comment by yatli** [comment]
Tested, everything works.

File: avr/cores/megacommand/MCL/FileBrowserPage.cpp [hunk 4]
```diff
@@ -101,6 +146,23 @@ void FileBrowserPage::init() {
     ((MCLEncoder *)encoders[1])->max = numEntries - 1;
   }
   DEBUG_PRINTLN(F("finished list files"));
+  ((MCLEncoder *)encoders[1])->max = numEntries - 1;
```
yatli, line 149: redundant.

avr/cores/megacommand/MCL/FileBrowserPage.cpp
```diff
@@ -407,25 +475,25 @@ bool FileBrowserPage::handleEvent(gui_event_t *event) {
...
 
     DEBUG_DUMP(temp_entry);
-    file.open(temp_entry, O_READ);
-
     // chdir to child
-    if (!select_dirs) {
-      if (file.isDirectory()) {
+    if (!show_samplemgr) {
+      file.open(temp_entry, O_READ);
+      // chdir to child
```
yatli, line 497: redundant

avr/cores/megacommand/make_and_upload.sh
```diff
@@ -6,3 +6,6 @@ if [ $? -eq 0 ]; then
   echo RAM USAGE: $(${AVR_DIR}/bin/avr-size main.elf | grep main | awk '{ print $2 + $3}')
   ${AVR_DIR}/bin/avrdude -C${AVR_DIR}/etc/avrdude.conf -v -V -patmega2560 -cwiring -P${DEV} -b115200 -D -Uflash:w:./main.hex
 fi
+
+make -j8
```
yatli, line 10: redundant.

jmamma: ??

yatli: Just marking a few redundant lines from the previous rebase. I'll clean it up when I have time.

yatli:
@jmamma I'd like to add a feature to the project file system -- sample slot descriptors.
The sample slot descriptors can be written in a file next to the grid files.
For every WAV transmission to MD, we update the descriptor file to track slot assignment. So when we move to another project, overwrite the sample slots, and then come back, we have a way to restore the samples for the current project.
For the ease of maintenance, we can reuse plain text ".SPK" file format which I already built for bulk-loading a sample pack. TODO: need to generalize .SPK to take absolute paths, and skip empty lines.
A sample slot descriptor file, therefore, contains 48 lines, each line pointing to a WAV file on the SD card. An empty line means the current project has not yet designated any samples to the slot.

Another advantage of this approach is, a new project can reuse the sample slot assignment of another project by simply going to the Sample Manager page, and load the .SPK file beside that project :)

jmamma:
So basically a SampleBank per project.

For every WAV transmission to MD, we update the descriptor file to track slot assignment. 

Yes, and someway of tracking/syncing the current Bank on the MD.

jmamma:
Could we implement some sort of fast hash on the MD side. and compare with local files...

yatli:
Won't match. MD samples are processed.

jmamma:
How big is the binary for this branch?

yatli:
The only workaround on the MD side, as I figured a while ago, is to examine the slot name...
In my current implementation it's the first 2 and last 2 chars of the original file name combined.

yatli:
One moment, I'll sync the Makefile changes to Arduino build.

yatli:
On Windows:
```
Sketch uses 234152 bytes (92%) of program storage space. Maximum is 253952 bytes.
Global variables use 49644 bytes (87%) of dynamic memory, leaving 7187 bytes for local variables. Maximum is 56831 bytes.
Low memory available, stability problems may occur.
```

yatli:
After removing references to `sprintf`:

```
Sketch uses 232730 bytes (91%) of program storage space. Maximum is 253952 bytes.
Global variables use 49650 bytes (87%) of dynamic memory, leaving 7181 bytes for local variables. Maximum is 56831 bytes.
Low memory available, stability problems may occur.
```

jmamma:
sprintf ? where is that being used?

yatli:
https://github.com/jmamma/MIDICtrl20_MegaCommand/pull/104/commits/ff06a9d0d17e91729500d6a461dabb8c3b7fe040

My bad. I naively thought that replacing 3 calls with 1 would reduce size :p

jmamma:
avr newb :grin:

jmamma:
We need to go through the firmware and replace instances of 

m_str* 
with standard library variants where they exist.

ie:
```
./MCL/Wav.cpp:  m_strncpy(&filename, file_name, 16);
./MCL/Wav.cpp:  m_strncpy(&filename, new_name, 16);
./MCL/FXPage.cpp:    m_strncpy_p(str, param_name, 4);
./MCL/FXPage.cpp:    m_strncpy_p(str, param_name, 4);
```

```
./GUI/GUI.cpp:  m_strncpy(data + idx, str, len);
./GUI/GUI.cpp:  m_strncpy(data + idx, str, m_strlen(str) - 1);
./GUI/GUI.cpp:  m_strncpy(data + idx, str, sizeof(lines[0].data) - idx);
```

yatli:
No much gain, but better than nothing:
```
Sketch uses 232418 bytes (91%) of program storage space. Maximum is 253952 bytes.
Global variables use 49650 bytes (87%) of dynamic memory, leaving 7181 bytes for local variables. Maximum is 56831 bytes.
Low memory available, stability problems may occur.
```

yatli:
Can you help double-check https://github.com/jmamma/MIDICtrl20_MegaCommand/pull/104/commits/87ae77d7f1ef77f5d5cc3c77a9c27565c35c6ac9

yatli:
Got to break out of this session. Preparing a talk.

yatli:
Rebased to latest dev.

yatli:
```
avrdude: 237280 bytes of flash written
```

Starting to have weird problems like the dreadful "SD CARD ERROR :-(", garbled file menu etc.

jmamma:
My binary **from dev branch** with debug on is 243918 and i'm not noticing any issues.

I'm wondering what's the cause of this issue, because previously I saw same problem as you when I went past 240K

PR #111, opened 2020-04-03
Wav editor

jmamma:
@yatli It works!

Performance is not too bad on 2second sample.

wav_test

IMG_2471
IMG_2470


jmamma:
Next step is add proper zoom functionality. Currently it just reduces visible samples by powers of 2, when rotating encoder0.

The plan will be use encoders 0 and 1 to create selection range (start sample, end sample). Press button to zoom. Press shift 2 to set loop points ?

jmamma:
Also, the peak search functionality only supports 16bit files currently.

Should we add support for 24bit?

(it means changing types from int16_t to int32_t. there will likely be performance penalties here)

yatli:
Wow.

jmamma:
24bit is a pain.

The problem is converting between 24bit sample (3byte stream) and int32_t

Negative numbers are stored as 2's complement. 

My approach in bool Wav::apply_gain(float gain, uint8_t channel) {
seems a little convoluted.

```
      int32_t sample_val;

      for (uint8_t b = 0; b < bytes_per_word; b++) {
        sample_val |= ((int32_t)buffer[byte_count]) << (b * 8);
        byte_count++;
      }
      // For signed formats, we need to remove the "sign" bit before
      // performing multiplication

      if (header.fmt.bitRate > 8) {
        bool is_signed = ((uint32_t)1 << header.fmt.bitRate - 1) & sample_val;
   
       if (is_signed) {
          // Signed numbers are stored as 2's complement :'-(
          // To perform any calculations we must first convert them to decimal
          // We also need to shake off any unwanted higher bits if we need to
          // perform a comparator operation.

          sample_val = ~(sample_val - 1);
          sample_val = sample_val << (32 - header.fmt.bitRate);
          sample_val = sample_val >> (32 - header.fmt.bitRate);
          sample_val &= ~((uint32_t)1 << (header.fmt.bitRate - 1));
        }
```

https://stackoverflow.com/questions/35876000/converting-24-bit-integer-2s-complement-to-32-bit-integer-in-c

Suggests if we move bytes so that they fill the upper bytes of int32_t and then shift down, the compiler will sign extend automatically, eliminating the need to convert from 2's complement.

---

I wanted to avoid bit shifting, to improve performance. 
May need to write separate code for handling 16bit /24bit streams, to gain back the efficiency in 16bit mode.

jmamma:
This is the code to read 24bit sample value and store in int32_t data.

```
      uint8_t word_offset = 4 - sample_size;
      //Move byte stream in to 32bit MSBs.
       for (uint8_t b = 0; b < sample_size; b++) {
         ((uint8_t*)&sample_val)[b + word_offset] = buffer[byte_count++];
       }
       //Down shift to preserve sign.
       sample_val = sample_val >> (word_offset * 8);
```
@yatli 
Performance penalty feels around 2-4x (for 16bit wav). Might be worth coding separately for 16bit. Thoughts?

yatli:
Where can I get those nice knobs? Haven't got time to paint the metal chasis, so I figure I may print a few "official" cases for nice-looking builds.

jmamma:
Those are elektron MD knobs. You can send them a support email asking to buy some replacement knobs for your MD. They're custom parts.

yatli:
Bad rendering file:
[Bd1.wav.zip](https://github.com/jmamma/MIDICtrl20_MegaCommand/files/4433050/Bd1.wav.zip)

Audacity rendering:
image

MC:
微信图片_20200405143711

jmamma:
haven't tested stereo yet. 

working on a few things then ill circle back.

yatli:
ah, I see :)

jmamma:
How does that look?
IMG_2474

jmamma:
Things to do:

24bit is not working.

Add menu, 
Menu item: switching draw_mode between left right and stereo view.
Menu item: add loop points to wave
Menu item: cut, copy, paste.

Zero crossing detection?

jmamma:
http://avr.2057.n7.nabble.com/24-bits-u-ints-td22397.html

24bit data type is supported by avr-gcc.

It's just not part of std inttypes.h

yatli:
That's cool, didn't know that!

Sent from Outlook Mobile<https://aka.ms/blhgte>

________________________________

jmamma:
bad news. ~chewing up 15KB .~

Might not be correct. will verify in the morning.

yatli:
> chewing up 15KB

In that case we might want to roll our own as well. We don't need advanced arithmetic anyway. If anything, we just promote to 32bit.

yatli:
That stereo wav looks fantastic!!

jmamma:
> In that case we might want to roll our own as well. We don't need advanced arithmetic anyway. If anything, we just promote to 32bit.

It was an error on my part. No noticeable size difference using the __int24 type.

jmamma:
https://www.cs.cmu.edu/~rbd/papers/cmj-float-to-int.html

need to check this ^

jmamma:
@yatli 
wav_editor is ready for a menu, and the ability to edit loop points.

jmamma:
@yatli added draw loop points.

"
Start
The start value specifies the byte offset into the waveform data of the first sample to be played in the loop.

End
The end value specifies the byte offset into the waveform data of the last sample to be played in the loop.
"

I've got

```
410   uint8_t i = 0;
411   uint8_t byte_rate = wav_file.header.fmt.numChannels * (wav_file.header.fmt.bitRate / 8);
412 
413   for (i = 0; i < wav_file.header.smpl.cSampleLoops; i++) {
414   
415     uint32_t loop_start = wav_file.header.smpl.loops[i].dwStart / byte_rate;
416     uint32_t loop_end = wav_file.header.smpl.loops[i].dwEnd / byte_rate;
```

Loop points don't look right unless I set byte_rate =1.

fyi: loop_start and loop_end are sample offset, not byte offset. 

---

@yatli is there an error in the spec?

Checking against Wavosaur the dwStart and dwEnd are treated as sample offsets, not byte offsets.

yatli:
I believe it's wrong spec. My first attempt to translate wav loop points to sds (following the spec) resulted in incorrect positions.

yatli:
Let's consolidate this PR into #104

PR #112, opened 2020-05-04
Chromatic Page Bug Fixes.

jmamma:
It should fix the following:

    Note to PTC mapping was incorrect. (Wrong note displayed on keyboard, machine base ptc not taken into consideration)
    Arpeggiator and Scales broken when using EXT MIDI. Fixed.
    Ext MIDI now responds from NOTE C2 and upwards. Scales are mapped across each note of keyboard and root is at C2.
    Poly config saved when exiting poly select page.
    Notes out of machine range, now result in silence, as opposed to playing highest note.
    Some tweaks to the arpeggiator GUI.

---

Also, I've noticed that MidiNotes.h definitions are incorrect.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,21 @@
+MCL 2.62 08/05/202
+
+Changes:
+
+  - Add transpose parameter to Chromatic Page. Accessible from Track Select Menu.
+
+Bug Fixes:
+
+  - Chromatic Page
+    * Note to PTC mapping was incorrect. (Wrong note displayed on keyboard,
+	machine base pitch not taken into consideration)
+    * Arpeggiator and Scales broken when using EXT MIDI. Fixed.
+    * Ext MIDI now responds from NOTE C2 and upwards. Scales are mapped across each note of keyboard and root is at C2.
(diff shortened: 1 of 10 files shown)


jmamma:
Just an aside note. I found it conceptually difficult to map a less than 12 note scale to a 12 x N note keyboard.

Had a quick look at how ableton does it, they use 'padded/repeated' notes so that scales are always 12 notes in length.

Our approach doesn't use padded notes, so the root note of the scale is different at each octave.
This is only a problem on a MIDI keyboard. On the MD trig interface the scale mapping is not a problem because we only have 16 keys.

jmamma:
Change octave/detune control to root note control ?

Currently all our scales are in the key of C.

yatli:
Very busy until the beginning of June. I may not be able to put chunky time on MC/MD (or music making in general) until then.

jmamma:
> 
> 
> Very busy until the beginning of June. I may not be able to put chunky time on MC/MD (or music making in general) until then.

All good. I have plenty to keep me busy here.

PR #116, opened 2020-06-04
Track Scales / Resolutions

jmamma:
Started adding support for track speed multipliers:

1x
2x
3/4x
2/3x
1/2x
1/4x
1/8x

These are now working for the MD.

However following needs doing:

- [x] Fix uTiming page for various scale modes. Timing is scaled according.
- [x] Fix chain mode for scale modes. Ie next_transition[] 
- [x] Add this scale support to A4/Ext tracks
- [x] Fix live record - [ ] Needs test
- [ ] Arpeggiator should be able to work with track scale. ie triplets.
- [x] Fix and test MD Merge (make sure correct scale is imported)
- [ ] Add ability to change scale of all tracks, move to pop up menu ?

### DIFF
--- a/avr/cores/megacommand/CommonTools/helpers.h
+++ b/avr/cores/megacommand/CommonTools/helpers.h
@@ -92,6 +92,16 @@ extern const uint32_t _bvmasks32[];
 #define IS_BIT_SET64(target, bit)           IS_BIT_SET_RADDR(target, bit)
 #define IS_BIT_CLEAR64(target, bit)         IS_BIT_CLEAR_RADDR(target, bit)
 
+/** 128-bit macros. **/
+#define SET_BIT128(target, bit)              SET_BIT_RADDR(target, bit)
+#define CLEAR_BIT128(target, bit)            CLEAR_BIT_RADDR(target, bit)
+#define TOGGLE_BIT128(target, bit)           TOGGLE_BIT_RADDR(target, bit)
+#define IS_BIT_SET128(target, bit)           IS_BIT_SET_RADDR(target, bit)
+#define IS_BIT_CLEAR128(target, bit)         IS_BIT_CLEAR_RADDR(target, bit)
+
+
+
+
 /** @} */
 
 /**
--- a/avr/cores/megacommand/MCL/A4Track.cpp
+++ b/avr/cores/megacommand/MCL/A4Track.cpp
@@ -5,7 +5,7 @@
 
 void A4Track::load_seq_data(int tracknumber) {
 #ifdef EXT_TRACKS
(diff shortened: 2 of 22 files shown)


jmamma:
I've merged this in to DEV.

@yatli if you are able to test...

jmamma:
fyi: slots saved in at this version, won't have a scale configured if you downgrade.

yatli:
unzipping my gears...

jmamma:
about time..

just sat down to do some coding.

- [ ] Pattern Speed / Length / Swing popup menu <- Low hanging fruit.
- [ ] Implement separate swing parameter. indepenedent of micro timing. 
- [ ] Or implement slides. (still trying to work out how to do this without division. )

MDTrackSeqData.h has uint8_t reserved[4]; which could be used for a uint64_t slide bit mask, or independent uint8_t swing control.

RAM usage currently:

34070 bytes (59%) of dynamic memory, leaving 22761 bytes for local variables.

ROM usage:

211144 bytes.

yatli:
merged dev into sysex_file branch. In step edit, uTim is jumping between +-1.

yatli:
> Or implement slides. (still trying to work out how to do this without division. )

Lookup tables?

int delta = tgt - cur;
int delta_delta = slide_lut[delta] >> speed_factor;
tgt += delta_delta;

jmamma:
> merged dev into sysex_file branch. In step edit, uTim is jumping between +-1.

I'm not seeing this here in dev branch. Can you help me narrow this down?

jmamma:
found the problem. late night refactoring. missing return statement.

jmamma:
> > Or implement slides. (still trying to work out how to do this without division. )
> 
> Lookup tables?
> 
> int delta = tgt - cur;
> int delta_delta = slide_lut[delta] >> speed_factor;
> tgt += delta_delta;

On the MD, speed factor is determined by number of steps between tgt and cur

It's linear interpolation between 2 steps.

jmamma:
i'm wondering if we can implement Bresenham's algorithm to do the slide interpolation .

--

Adafruit_GFX::writeLine() is a very light weight implementation. we can use this.
Now just need a quick way of finding tgt step.

jmamma:
> The default behaviour for parameter locks is that the value changes exactly when the step is played and not in between. By using slides, you can have the value of a parameter locked on one step slide gradually to the value on the next. This behaviour is controlled by the slide pattern or the slide tracks.

> For a parameter to slide it needs to be parameter locked at two separate steps. If a slide trig is set to the same step as there is a parameter lock, the parameter will slide to the next parameter locked value. The speed of the slide is relative to the current tempo and the slide is completed when the next position with a drum trig is reached.

PR #117, opened 2020-06-06
Transitions for 3/4 and 3/2 track speeds.

jmamma:
Will likely need to rethink how track transitions works.
We currently calculate next transition in 16th step resolution.

This won't work for 3/2 and 3/4 tracks because they run at 0.75 and 1.5 speed.

For example:

Pattern length is 16 steps. At 3/4 speed, the next transition will occur on (1/0.75) * 16 == 21.333 steps.

Sequencer will go out of sync...

---

Proposed solution. Each track to have a transition_offset, measured in mod12 ticks. Transition start can be shifted by 0 -> 12 ticks

### DIFF
--- a/avr/cores/megacommand/MCL/ExtSeqTrack.cpp
+++ b/avr/cores/megacommand/MCL/ExtSeqTrack.cpp
@@ -61,8 +61,9 @@ void ExtSeqTrack::seq() {
   if (mute_until_start) {
 
     if (clock_diff(MidiClock.div16th_counter, start_step) == 0) {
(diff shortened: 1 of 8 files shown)


jmamma:
transition_offsets added.

Still going out of sync... with MD sequencer on transitions.

jmamma:
Transitions for 
3/4x 3/2x now working via track offset method described above.

Question: Should the offset be reset to 0 if transitioning back to a standard 1x, 2x scale.
Answer: MD does not reset clock offset, if switching between patterns of different scales.

PR #118, opened 2020-06-09
Slides

jmamma:
Just coded up the Bresenham algo, based on the Adafruit_GFX::writeLine implementation.

The basic idea is that once a track is triggered, it is flagged to have slide data recalculated.
This recalculation occurs at the end of the sequencing loop to prevent latency.

The slide data recalcaulation involves determining x1 - x0, by calling find_next_lock(uint8_t step, uint8_t param) . This will find the next matching lock of the same param type.

find_next_lock(uint8_t step, uint8_t param) worst case performance is 64 itterations.

Once x1 -x0 is determined, the parameters for Bresenham algorithm are calculated and cached in SlideData locks_slide_data[4]; 

On subsequent calls of seq() (every 192th div), send_slides() will be called and will run through the the Bresenham routine, updating and sending the slide parameter value.

### DIFF
--- a/avr/cores/megacommand/MCL/ExtSeqTrack.cpp
+++ b/avr/cores/megacommand/MCL/ExtSeqTrack.cpp
@@ -37,7 +37,7 @@ float ExtSeqTrack::get_speed_multiplier(uint8_t speed) {
 void ExtSeqTrack::set_speed(uint8_t _speed) {
   uint8_t old_speed = speed;
   float mult = get_speed_multiplier(_speed) / get_speed_multiplier(old_speed);
-  for (uint8_t i = 0; i < 128; i++) {
+  for (uint8_t i = 0; i < NUM_EXT_STEPS; i++) {
     timing[i] = round(mult * (float)timing[i]);
   }
   speed = _speed;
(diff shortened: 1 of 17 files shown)


jmamma:
Semi working.


Although
-  uint8_t reserved[4];

^ only have 4 bytes in reserve. This is not enough for uint64_t slidemask.

jmamma:
Implementation is working. 

I've just tested performance, it's good!

12 Tracks with 4 slides per track running continuously is about the limit for the MD.
at 16 Tracks with 4 slides, the GUI starts to lockup.

The MC is not even breaking a sweat.

jmamma:
utiming is being set to 0 value in SeqStepPage. 
Related to recent changes in dev, Investigate.

jmamma:
Slide functionality is complete.

Just need to work out how to integrate it in to MCL.
The MDSeqTrackData.h spec will need to change to incorporate a 
uint64_t slidemask;

jmamma:
@yatli merged in to dev.

have  a play.
slide mask can be toggled from StepPage shift menu.

fyi: slide mask is not persistent. we need to change the project spec for this -> 3.0

PR #123, opened 2020-07-10
seqpage/pageselectpage TI led

yatli:
Requires https://github.com/yatli/mdhack/pull/20

Demonstrates two TI trigled modes: normal and rec.

PageSelectPage uses the normal mode to show the current selected page. Playing notes will be displayed on the TI.
SeqPage uses the rec mode to show MCL mask data plus the current playing cursor.

### DIFF
--- a/avr/cores/megacommand/MCL/GridSavePage.cpp
+++ b/avr/cores/megacommand/MCL/GridSavePage.cpp
@@ -145,6 +145,7 @@ bool GridSavePage::handleEvent(gui_event_t *event) {
   }
 
   if (note_interface.is_event(event)) {
+    trig_interface.send_md_leds();
     if (note_interface.notes_all_off()) {
       if (BUTTON_DOWN(Buttons.BUTTON2)) {
         return true;
--- a/avr/cores/megacommand/MCL/GridWritePage.cpp
+++ b/avr/cores/megacommand/MCL/GridWritePage.cpp
@@ -129,6 +129,7 @@ bool GridWritePage::handleEvent(gui_event_t *event) {
   DEBUG_DUMP(event->source);
   if (note_interface.is_event(event)) {
     DEBUG_PRINTLN("note event");
+    trig_interface.send_md_leds();
     if (note_interface.notes_all_off()) {
       DEBUG_PRINTLN("notes all off");
       if (BUTTON_DOWN(Buttons.BUTTON2)) {
--- a/avr/cores/megacommand/MCL/MCLGUI.cpp
+++ b/avr/cores/megacommand/MCL/MCLGUI.cpp
@@ -608,7 +608,7 @@ void MCLGUI::draw_trigs(uint8_t x, uint8_t y, uint8_t offset,
           oled_display.fillRect(x + 1, y + 1, seq_w - 1, trig_h - 1, WHITE);
         }
 
-    }
+      }
     }
 
     x += seq_w + 1;
--- a/avr/cores/megacommand/MCL/PageSelectPage.cpp
+++ b/avr/cores/megacommand/MCL/PageSelectPage.cpp
@@ -136,6 +136,8 @@ void PageSelectPage::init() {
   classic_display = false;
 #endif
   loop_init = true;
+  // clear trigled so it's always sent on first run
+  trigled_mask = 0;
   display();
 }
 
@@ -177,7 +179,7 @@ void PageSelectPage::cleanup() {
   MDSysexListener.removeOnKitMessageCallback(&kit_cb);
 #endif
   note_interface.init_notes();
-  MD.set_trigleds(0, TRIGLED_EXCLUSIVE);
+  MD.set_trigleds(0, TRIGLED_OVERLAY);
 }
 
 uint8_t PageSelectPage::get_nextpage_down() {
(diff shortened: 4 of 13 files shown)


jmamma:
Just testing this out now.

The latency is much improved. ~still a few ms delay on the MD side. But good enough for now.~

There is a bug, if you enable REC on the MD. you will input step sequencer data on the MD sequencer, without seeing it visually.

jmamma:
Can we re-name REC mode to something else. STEPINPUT perhaps ?

jmamma:
~In "REC" mode we need to show the current selected trigs, in addition to the trig mask.~
Because the SET_BIT64() is set on trig release, there is perceived lag on the MD side.

Edit: i think we just need to mimic the MD step edit behaviour more closely.
If the note has no trig, then set trig regardless of "hold_time"

jmamma:
~Trig Latency appears to also be present in the MD's standard step edit mode.~

I think this is just my worn out MD triggers. Tested against my 'backup' MD and the response is the same.

jmamma:
9725f5d

Step input should feel more responsive now. Step will be set if empty on press.

jmamma:
We should disable MD step edit mode if in MCL step edit mode.

yatli:
> There is a bug, if you enable REC on the MD. you will input step sequencer data on the MD sequencer, without seeing it visually.

This is because currently our TI hack only handles the non rec mode path.
We can setup a MCL presence flag in the FW, and disable the MD step edit accordingly.

Then we can use the MD rec button to enter the MCL step edit mode?

jmamma:
> Then we can use the MD rec button to enter the MCL step edit mode?

I thought about this. but it might not be desirable. For instance I regularly find myself in the MD step edit mode with the MC attached.

--

Just noticed that MUTE step leds are now broken too when in MCL step edit. Same issue as above.

yatli:
Makes sense. I'll think about the precedence and override rules.

yatli:
@jmamma 

```c++
  // Do not show stepedit trigled if:
  // - trigger interface is not activated
  // - MD rec mode is activated
  // - MD bank selection is activated
  // - MD seq edit (mute, slide etc.) is activated
  if (BankButtonPressIndexForPopup >= 0  ||
      md_menu_cmd.md_menu_cmd5 != RecordDeactivated || 
      md_menu_cmd.md_menu_cmd4 != 0 ||
      !trigger_interface_en){
    return LEDS_dynamic[i] ^ LEDS_static[i];
  }
```

jmamma:
> @jmamma
> 
> ```c++
>   // Do not show stepedit trigled if:
>   // - trigger interface is not activated
>   // - MD rec mode is activated
>   // - MD bank selection is activated
>   // - MD seq edit (mute, slide etc.) is activated
>   if (BankButtonPressIndexForPopup >= 0  ||
>       md_menu_cmd.md_menu_cmd5 != RecordDeactivated || 
>       md_menu_cmd.md_menu_cmd4 != 0 ||
>       !trigger_interface_en){
>     return LEDS_dynamic[i] ^ LEDS_static[i];
>   }
> ```

does this work?

yatli:
https://github.com/yatli/mdhack/pull/20/commits/727de6923811a3930e10034fdef04b293336e6a7

Try the latest one.
When MD rec mode is active, the LEDs display the internal sequencer state (same for mute/accent etc.)

yatli:
It even works with bank selection. When the popup times out, the stepedit TI comes back online :D

yatli:
Only issue, MD does not know the MC step cursor position. I'm currently using the internal sequencer step cursor.

Can we sync the SCALE SETUP over from MC?

yatli:
Tried more patterns with the PageSelectPage TI led index indicator -- maybe we should go exclusive instead of overlay

yatli:
Overlay leds for chain/write indicator would be fun. I can then see which tracks are playing notes and chain those.

jmamma:
> Only issue, MD does not know the MC step cursor position. I'm currently using the internal sequencer step cursor.
> 
> Can we sync the SCALE SETUP over from MC?

Yeah that would be awesome. I found myself trying to change the current "page" of the sequence using the scale button on the MD. Only to remember i had to do in the MC.

jmamma:
this is really good now.

jmamma:
We need another mode. to hide the sequencer events.

For example, it's hard to see the current page select, when the sequencer is running

---

And we need a mode, to prevent XOR from sequencer events. 

So if you hold down trigs, the sequencer events don't XOR the held trigs.

test commit below to see behaviour:

yatli:
see https://github.com/jmamma/MIDICtrl20_MegaCommand/pull/123/commits/3e3f9426533e0481c9e23cdb7f73a33751130ac4

yatli:
PageSelectPage not sending the first leds update
edit: fixed.

jmamma:
> 
> 
> see [3e3f942](https://github.com/jmamma/MIDICtrl20_MegaCommand/commit/3e3f9426533e0481c9e23cdb7f73a33751130ac4)

need the patch on the MD side.

yatli:
Forgot to commit...

https://github.com/yatli/mdhack/pull/20/commits/6bb64b040476a41c2b4de934630dc780b9c7d587

jmamma:
> 
> 
> Forgot to commit...
> 
> [yatli/mdhack@6bb64b0](https://github.com/yatli/mdhack/commit/6bb64b040476a41c2b4de934630dc780b9c7d587)

yep that's perfect.

--

Now just the issue of the step indicator.
This will be broken for track speeds and different track lengths.

yatli:
merging this one for cleaner code review on https://github.com/jmamma/MIDICtrl20_MegaCommand/pull/124

PR #124, opened 2020-07-30
Piano roll

jmamma:
First approach is to create a FOV system. Similar to WavEditor.

We have 4 note patterns. 

For each note pattern,
  Scan entire length for NOTE_ONs.
    For each NOTE_ON determine start and end positions. (requires scanning for corresponding NOTE_OFF with potential for wrap around.)
     Determine if note falls within FOV and should it be drawn.


jmamma:
@yatli layed out the basic routine for drawing the piano roll.

Feel free to jump in if you like, otherwise i'll pick this up sometime tomorrow.

jmamma:
oh yeah. remind me to rename the timing_mid variable. used all over the sequencer code.
A better name is ticks_per_step. this changes depending on track speed.

For example, @ 1x timing_mid =12 (div192 ticks per step)
                           2x timing_mid = 6 etc...

yatli:
I'm working on another round of size optimization hunting. If you have time on this one please go ahead, I'll catch up later.

yatli:
what does "wrap around" mean here?

jmamma:
if the note_on event is recorded before the end of the sequence but the note_off is recorded after the loop

For example, pattern length 16. You could have:

note_on step 14.. note_off step 5

jmamma:
draw_line (and other draw functions) are not ranged checked with respect to the buffer.

For example, if you drawling(127 + x, 0, 127 +x, 32); you'll be writing outside of the buffer array.

This is kind of dangerous.

jmamma:
vertical scroll can be used as an interactive cursor, much like the grid page. when the cursor reaches the end of the screen top/bottom scroll up down

jmamma:
@yatli piano roll draw routine should be working now. 

Encoder1 translate x (ticks)
Encoder2: translate y (notes)
Encoder4: horizontal zoom.

The number of notes on screen can be adjusted with the const ~fov_h~ fov_notes.

Inputting notes is completely broken. So to test, I record a sequence on the A4 using the chromatic page, then flip back to ExtStep page to see the result. Make sure you record within vertical range of the visible notes, or use y translate.

--

Next on the the todo list

- [x] Interactive cursor (copy code from grid)
- [x] Vertical chromatic keyboard to be drawn in line with notes.

jmamma:
Added chromatic keyboard, grid with 16th note division and verticle projection of notes.

Interface is feeling a bit cramped now. 

Also fov_w needs to be 96 pixels to scale nicely. The chromatic keyboard takes up 4, so that 100 pixels in total.
We can get away with it if we compress the side menu drawn by SeqPage::display() slightly.

yatli:
Kind of embarrassing question... How do I go to the ext tracks in step edit mode now?

jmamma:
> Kind of embarrassing question... How do I go to the ext tracks in step edit mode now?

lol. you coded this part didnt u?

Press save from StepEdit page.

yatli:
But then I'm stuck at channel 01 and not able to move around = =!

jmamma:
channel select is done on the A4 side.

Do you have  your A4 connected. press keyboar on A4 and MCL will midi learn channel

yatli:
Ah.

I got MD+MM connected at the moment. Totally forgot the A4 workflow..

jmamma:
ah you know what. previously the track shift menu had "TRACK SELECT" option.
That option is now disabled if MCL is configured to listen to MD track_select messages.

Oversight. forgot about the ext tracks.

jmamma:
Fix above ^

track shift menu is currently drawn before the piano roll  and will appear distorted, you can still access the options though.
this will be resolved when we "compress the side menu drawn by SeqPage::display() slightly"

jmamma:
vertical chromatic keyboard is probably not very useful?
hard to see which notes are which

yatli:
I guess we can further optimize the representation. Currently white keys are concrete and the blacks are just "now drawn"

jmamma:
So the next task is to create the interactive cursor (similar to how  the grid cursor behaves).

Then write the functions for drawing notes in the sequencer.
And integrating the trigger interface somehow ?Trig interface could be used to move the cursor around in divisons of 16?

I am liking the idea of being able to draw notes with just the MC and not use an attached device. this could prove fruitful as a standalone sequencer.

yatli:
Oh and, one nice thing about the matrix sequencer is that notes vertically outside the fov are drawn half width at the top/bottom.

jmamma:
> Oh and, one nice thing about the matrix sequencer is that notes outside the fov are drawn half width at the top/bottom.

vertical projection? i've added that in 
38dcb22 and width was reduced to half in e5356a8

Only draws a single projection across the bottom horizontal row of the display.

yatli:
Oh okay didn't catch that one : D

wrt TI, I think the vertical projection could just extend to the LEDs?

That makes the projection consistent, and the display will serve as a visual cue to the TI LEDs.

jmamma:
> Oh okay didn't catch that one : D
> 
> wrt TI, I think the vertical projection could just extend to the LEDs?
> 
> That makes the projection consistent, and the display will serve as a visual cue to the TI LEDs.

sure. if you want to implement this we can see how it looks/feels

yatli:
Originally I though the horizontal zoom is more quantized, like, bound to the speed.
We could use the save button as the pager.

jmamma:
My intention was to make the piano roll 'continous time'. The finest resolution is div192. This follows on from my work on the parameter slides, which also work at this timescale

The equation for zoom is.
```
fov_length = fov_zoom * timing_mid;
```
where fov_zoom = encoder2.cur

timing_mid = the number of div192/ticks per 16th note. 

So every increase/decrease in zoom, removes/adds a 16th note from the visible fov.

jmamma:
One last thing to think about...

```
 33 
 34 class ExtSeqTrackData {
 35 public:
 36   uint8_t length;
 37   uint8_t speed; // Resolution = 2 / ExtPatternResolution
 38   uint8_t reserved[NUM_EXT_NOTES];
 39   int8_t notes[NUM_EXT_NOTES][NUM_EXT_STEPS]; // 128 steps, up to 4 notes per step
 40   uint8_t locks[NUM_EXT_LOCKS][NUM_EXT_STEPS];
 41   uint8_t locks_params[NUM_EXT_LOCKS];
 42   uint64_t lock_masks[NUM_EXT_LOCKS];
 43 
 44   uint8_t conditional[NUM_EXT_STEPS];
 45   uint8_t timing[NUM_EXT_STEPS];
 46 }; 
 47 class ExtS
```

Currently there is a limitation that a single utiming value  is shared amongst all 4 notes of  the same step.

This means, that notes starting and ending on the same step must share that microtiming. So individual note length duration (A function of the micotiming value for note_start and note_end) is limited if notes coincide together.

I was considering converting the ```locks[NUM_EXT_LOCKS][NUM_EXT_STEPS];``` above to microtiming, to give us fully independent note lengths. 

And then converting timing[NUM_EXT_STEPS] to VELOCITY.

Then we could add parameter locks later, without compromising the project spec...

jmamma:
Cursor now controlled using encoders 1 + 2. when it reaches edge of fov, the fov is translated accordingly.

cursor position in div192 ticks is represented by these variables:

cur_x  
cur_y
cur_w

cur_w will be controlled by encoder3, I need to implement this next.

jmamma:
Tempted, to make encoders 3 + 4  width and height selection...

copy + paste notes? cut notes in half 😱 

--

piano roll is only one leap away from "chain mode, arranger view"

jmamma:
cur_w is now implemented. You can expand the cursor width using encoder 3 and it will translate the fov as it grows. 
I've also limited the piano roll fov bounds and zoom so it feels more contained now.

Cursor is now complete. 

We can now add functionality to draw notes.

--


~@yatli i'm trying to think of a clever way to distinguish the cursor from the drawn notes. I was thinking either a dotted drawRect pattern. Or somehow we need to implement a proper drawRect(x,y,w,h,INVERSE) function.~

jmamma:
Oh yeah, regarding quantization. 

My plan:

Pressing trig interface to move cursor to positions in the fov in divisions of 16.
When trig interface is held, changing the note_w will also be quantised. ie enable snap to grid if trig held.

jmamma:
added top projection. feels more intuitive.

jmamma:
break time.

yatli:
The piano roll overlaps with the left info panel.. Thoughts?

jmamma:
Yeah, we could forgo the piano roll or compress the SeqPage::display side menu a by a few pixels... 

The piano roll needs to be 96 pixels to scale nicely (excluding chromatic keyboard
)

jmamma:
Compressing the side menu will probably make it look cramped. particularly the "sequencer page" indicator.

jmamma:
drop the piano roll. save a few bytes.. ?

yatli:
> drop the piano roll

😱

yatli:
replace horizontal zoom with quantized viewport, align viewport to seq page?
I can try implement this one.

jmamma:
```replace horizontal zoom with quantized viewport, align viewport to seq page?```

what do you mean exactly?

--

also, I just realized if we resize the side menu globally, all other pages will look weird.

**Comment by yatli** [comment]
Say the pattern is 64 steps. Initially we start at zoom level 100%.
The seq page indicator is drawn as the normal 4 rectangle with current displayed page highlighted.
Instead of letting the cursor freely scroll the horizontal view, we use the seq page button to scroll between the 4 pages.
Cursor going beyond the last step will wrap around to step 1 of the next page.
We can switch zoom level at 25%, 50%, 100%, 200% etc. but not arbitrary values.

If we go for 200% zoom, the seq page indicator becomes 8 small squares. If we go 50% it becomes two longer rectangles. Switching to 25% causes the indicator becomes a single solid rectangle, meaning that we are at the lowest zoom level and everything is displayed, and there're no other seq pages to switch to.

Anything above 200% then we can draw the page indicator as a slide bar for fine-grained position display.

---

Of course, the tradeoff is that utiming cannot be displayed, and we must bring back the utiming menu.
I don't think this is all that bad though, because I really don't want to zoom in 100x to adjust the position of a note, zoom back to 1x and focus on the next note, zoom to 100x to adjust again. I'd rather have a complete view of the notes at 1x and be able to adjust utiming right there.

**Comment by jmamma** [comment]
> Say the pattern is 64 steps. Initially we start at zoom level 100%.
> The seq page indicator is drawn as the normal 4 rectangle with current displayed page highlighted.
> Instead of letting the cursor freely scroll the horizontal view, we use the seq page button to scroll between the 4 pages.
> Cursor going beyond the last step will wrap around to step 1 of the next page.
> We can switch zoom level at 25%, 50%, 100%, 200% etc. but not arbitrary values.

I think we should have a combination of both continuous. and paged scroll.

The zoom you're talking about should be a 1line change.

fov_length = fov_zoom * timing_mid;

where fov_zoom = number of steps you want to show on screen. currently fov_zoom = seq_param4.cur . but you could change scale that to be quantized.

**Comment by yatli** [comment]
image

The benefit of this layout is a strict mapping of 5 pixels per step and one empty pixel between two steps. We can therefore draw 2-pixel high notes and still able to figure out the boundaries of notes. A continuous note draws over the empty pixel, and two consecutive short notes don't. We don't have to go for 3-pixel high notes so that the boundaries can be drawn.

**Comment by jmamma** [comment]
If you want to see how 2 pixel notes looks (without spacing) you can change
fov_notes = 14 to give you 2 pixels per note .

**Comment by jmamma** [comment]
how does the 1 pixel spacing work at different zoom levels?

**Comment by yatli** [comment]
Ok I'm trying to look into the code now :)

FYI a 2-pixel wide chromatic keyboard mockup:

image

My neural network just completes the missing details for me..

**Comment by jmamma** [comment]
if you want better definition between adjacent notes, we could draw a rounded rect...

**Comment by yatli** [comment]
> how does the 1 pixel spacing work at different zoom levels?

This is zoom agnostic. Each step still takes 5 pixels and spacer takes 1. But one step is now twice/half the length etc.

> if you want better definition between adjacent notes, we could draw a rounded rect...

But that still requires h=3 right..

**Comment by jmamma** [comment]
brain cant compute, does this work with microtiming?

the current piano roll, can represent the exact length of notes.

**Comment by yatli** [comment]
Theoratically it does work with utiming. My argument is that given we have 96 width it's hard to exactly represent note length without heavy aliasing anyway...

**Comment by yatli** [comment]
A full mock with grid drawn:

Zoom=1x:

image

**Comment by yatli** [comment]
Zoom=2x, showing first half of the pattern above:

image

The page indicator becomes a slider, and each note takes double steps.
The page indicator can display both the page position and current zoom level (roughly).
The spacer still takes 1px width. But drawn at half the density.

The zoom level can be any integer value, like 1x, 2x, 3x, 4x, ... and also 0.5x, 0.33x, 0.25x ... 0.125x. There's no point going further than that. because the maximum length is at 128 (0.5x speed)

**Comment by jmamma** [comment]
you just want matrix sequencer :p

**Comment by jmamma** [comment]
i'm not convinced by your proposal. i'd need to compare the two designs side by side.

**Comment by jmamma** [comment]
matrix sequencer vs abl3t0n piano roll lol

**Comment by yatli** [comment]
> matrix sequencer vs abl3t0n piano roll lol

That's good for discussion to distinguish the two :D

**Comment by yatli** [comment]
btw I just tried fov_h=14 and it looks good to me. The chromatic keyboard suddenly starts to make sense with no change to the code. I guess it's just with h=3 notes there aren't enough space for my brain to imagine there's a vertical keyboard...

**Comment by yatli** [comment]
What's your plan on the seq page button if going with current design?

**Comment by jmamma** [comment]
> btw I just tried fov_h=14 and it looks good to me. 

looks too small to me. chromatic still doesnt look that great

**Comment by jmamma** [comment]
Save button = add note?


> 
> 
> What's your plan on the seq page button if going with current design?

could still be used to jump to pages by setting the piano roll offset

**Comment by jmamma** [comment]
> 
> 
> > btw I just tried fov_h=14 and it looks good to me.
> 
> looks too small to me. chromatic still doesnt look that great

looks okay if the grid is disabled. so maybe some tweaking if used at this resolution?

Chromatic keyboard can probably be improved slightly by drawing a solid  vertical white line over last pixel (so that black notes are 1 pixel shorter) (uncomment line 146. shift left 1 pixel)

**Comment by yatli** [comment]
yep, doing that now.
and keyboard_w can be lowered.

**Comment by yatli** [comment]
@jmamma is there a quick hack to support probing mnm? It always gets stuck at peering for a short period.

**Comment by jmamma** [comment]
Alright have a play. 

I'm going to adjust the sequencer code so that we can do micro timing per note, not just per step.

**Comment by jmamma** [comment]
> @jmamma is there a quick hack to support probing mnm? It always gets stuck at peering for a short period.

MNM not currently supported in active peering. You need to write a 'driver' for it.

Actually the MidiActivePeering.cpp code needs a  a re-write or re-think.

Originally it was supposed to identify any device via:

 ``` 
70     if (MidiUart.device.getBlockingId(DEVICE_MD, UART1_PORT,
 71                                       CALLBACK_TIMEOUT)) {
 72       DEBUG_PRINTLN("Midi ID: success");
```
But it turned out that elektron didnt implement the MidiID on the Analog 4.
http://www.somascape.org/midi/tech/spec.html#usx7E0601

--

MidiActivePeering.cpp is actually a good candidate for freeing up a few firmware bytes, by simplifying the code.

jmamma:
Port 2 is currently a bit of a hack, in that it only looks for the a4 (by requesting a global settings sysex dump).
If no dump is received after 3 attempts it falls back to general midi.

jmamma:
Line 192: MidiActivePeering.cpp

Replace a4_setup(); with
MidiUart2.device.set_id(DEVICE_MIDI);

That should set port 2 to general midi when activity is detected on the port. (activity is any midi message received)

yatli:
> by requesting a global settings sysex dump

I was just going to say that, and you've already got this done :D

yatli:
This is a bit like the Linux kernel driver probes. None of the devices seem to conform to a single standard (including those miserable ARM SoC situations where you cannot tell the presence of something unless the datasheet of the exact chip tells you so), so we can see all kinds of weird probing methods going on.

So I see the global request thing is totally clean :p

yatli:
@jmamma check 8f4da9f for "continuous page index indicator"

yatli:
@jmamma https://www.elektronauts.com/news/522

They got step recording now:

```
A new sequencing mode is available on the Digitakt and Digitones - Step Recording. This feature lets you input chords and notes into the sequencer step by step. Or use jump mode to skip to the next predetermined spot in the sequence via the Length parameter, making it even easier to control your melodies and build them a step at a time.
```

(envy)

jmamma:
TB 303 had that in 1984 😎

jmamma:
I’ll be back in the studio In a couple of hours. Needed to fetch my scope from the workshop

yatli:
> TB 303 had that in 1984 😎

Yeah those old Japanese sequencers.. Really creative.

I have a VL-Tone that does decoupled melody vs. rhythm -- you first "step record" the notes, then playback the piece and use the two "picker" keys to set the time offset for each note like playing right hand on a guitar or bass.

Good for a bad player like me.

jmamma:
Here’s one for you. We can use MD track encoder for transport /cursor control

yatli:
> Here’s one for you. We can use MD track encoder for transport /cursor control

But it's one dimension (?)

jmamma:
Just uploading the new patches now.

jmamma:
Too cramped with chromo.

I have an idea draw keyboard* on top of sequencer.
if cursor is left of centre draw keyboard on right. if cursor is right of centre draw keyboard on left.

Or we could have a fade in/fade out on the left. if the cursor approaches the keyboard fade it out.

jmamma:
i like the minimap btw.

yatli:
How about just a vertical minimap.
Display current cursor pitch in the info pane.

jmamma:
> Display current cursor pitch in the info pane.

Yes we can fit the Note on bottom Row. 
There's a uin8_t to note routine floating around on SeqStep page. 

```
159 MusicalNotes number_to_note;
160         uint8_t oct = notenum / 12 - 1;
161         uint8_t note = notenum - 12 * (notenum / 12);
162         strcpy(K, number_to_note.notes_upper[note]);
163         K[2] = oct + '0';
164         K[3] = 0;
```

> How about just a vertical minimap.

Can you elaborate?

yatli:
Just another projection from all the notes onto the full range, and the current viewport range.
Kind of like a text editor enhanced scroll bar showing "changed lines" with little green bars.

Not exactly sure if this will work though. w=2 is too cramped

jmamma:
I started refactoring the code for poly micro timing only to realise we would also need to allocate additional space for poly conditional notes.

Thoughts ?

yatli:
~~Even the huge A4 track cannot hold the data?~~

Guess I asked the wrong question.
`#define A4_TRACK_LEN (sizeof(GridTrack) + sizeof(ExtSeqTrackData) + sizeof(A4Sound))`

yatli:
We should make GridTrack a more generic container.

```
// GridTrack is always followed by track data + sound data
// So describe the components here.
uint16_t device_id;  // identifies sound data; prevent sending parameters to a wrong device
// mono: MD-style track
// multi: A4/MnM-style multi voice. If using multiple voices, can leverage utiming/conditional space from other tracks (?)
// poly: true poly voice. none of such devices have been connected to MC as of now. 
uint16_t track_type; // TRACK_TYPE_MONO, TRACK_TYPE_MULTI, TRACK_TYPE_POLY, TRACK_TYPE_AUTOMATION
```

Problem for POLY: suppose we make individual utiming/conditiion for notes, then what about the locks?

Problem for ext (multi/poly): there are 128 steps but the lock masks are 64 bits wide.

jmamma:
> Guess I asked the wrong question.
> `#define A4_TRACK_LEN (sizeof(GridTrack) + sizeof(ExtSeqTrackData) + sizeof(A4Sound))`

./Grid.h:#define GRID_SLOT_BYTES 4096

I've pre-allocated a max of 4096 bytes per slot within the current project layout. So we can increase the current track sizes. The issues are:

1. Byte layout / Supporting different track versions
2. RAM consumption
3. Loading and caching time.
4. Sequencer performance

1) I have a mechanism for versioning. To distinguish a new track version we can shift the track speed by an offset, then on load check to see if the offset is in range. If the offset is not in range perform conversion from old format to new format.

2) We still have plenty of RAM but should allocate it conservatively.
I'm considering increase MD locks to 6 or 8 (subject to  3. and 4.)

3) This will effect performance in chain mode. Every increase in 512 bytes is another read to the sd card. Need to make sure we can cache 20 tracks of data before the shortest next transition (2 * 16th notes).
 
4) If we add more locks, we increase the number of loops per track in the sequencer code. Both parameter locks and slides have number of loop iterations = NUM_LOCKS

yatli:
> // multi: A4/MnM-style multi voice. If using multiple voices, can leverage utiming/conditional space from other tracks (?)

What do you think of this?

jmamma:
memory regions need to be kept isolated so that they can be swapped in and out .
You'd have to tag each data point with the corresponding track_id otherwise.

jmamma:
I'm just finishing up a proposal/implementation for track conversion going forward... 

I'll push up a separate PR soon.

yatli:
@jmamma just back home from 2nd trip.
Looks like you've got a ton of work done!!

jmamma:
😓 yes

jmamma:
That's it for me tonight.

The last commit solves the overlapping note problem, by deleting conflicting notes before adding new notes.

But this breaks adjacent notes. Will need to look at this tomorrow.

avr/cores/megacommand/MCL/SeqExtStepPage.cpp
```diff
@@ -534,8 +533,9 @@ void SeqExtStepPage::del_note() {
         uint16_t note_end = j * timing_mid + ev_j.micro_timing - timing_mid;
 
         if ((note_start <= cur_x + cur_w) && (note_end >= cur_x)) {
-          active_track.remove_event(note_idx_on);
+          DEBUG_DUMP("deleteing");
           active_track.remove_event(note_idx_off);
```
jmamma: @yatli  
can you check this.

The note_idx_off event does not get deleted with the ordering (on -> off). Probably a bug in remove_event.

yatli:
@jmamma check out the latest commits. I love how this X-Y input scheme enlightened me to write something I wouldn't write before.

100x better than Elektron sequencer for chords

yatli:
Hah. I feel like being Douglas Engelbart doing [The Mother of All Demos](https://www.youtube.com/watch?v=JQ8ZiT1sn88)

yatli:
> But this breaks adjacent notes

Observed this.

Also saving isn't working for me.

jmamma:
@yatli 

At the risk of assigning us even more work to do:

class MDPolyTrack : ExtSeqTrack {

}

Then allow for piano roll editing of MD poly tracks. (Mono mode defaults to standard MD step edit, poly uses piano roll)

Somehow Grid/Sequencer would need to support both MDSeqTrack and MDPolyTrack types.

avr/cores/megacommand/MCL/SeqExtStepPage.cpp
yatli: When you delete note_idx_on, the index for off event changes 😅

yatli: but you have worked out an elegant solution 👍

yatli:
> oh yeah. remind me to rename the timing_mid variable. used all over the sequencer code.
> A better name is ticks_per_step. this changes depending on track speed.
> 
> For example, @ 1x timing_mid =12 (div192 ticks per step)
> 2x timing_mid = 6 etc...

Ding.

yatli:
About MD poly track, I think it's better to group multiple MD mono tracks rather than use a ext track.

1. MD poly has to span over multiple tracks.
2. Chroma recording also works like that.
3. Mixing ext and MD mono tracks is a huge challenge from memory management perspective.

avr/cores/megacommand/MCL/SeqExtStepPage.cpp
jmamma: lol an accidental solution.

jmamma:
**PianoRoll Todo:**

The encoder buttons could be used to switch between fine and coarse movements / note width changes. 

I will add multi note support for add_note() when notes are held on UART2.

Shift Menu items to be added:
- Pattern Length
- Note conditional
- Edit mode (select notes or input note)
- Cut/copy/paste notes
- Toggle piano/roll vs event editor (parameter locks/slides etc)

Other:
- PageSelect: REC page to be removed and replaced with Piano Roll
- SeqStepEdit Save button to enable live record of both trigs and parameter locks.
- We could do the same for ExtSeqStepEdit. Save button enables live record. Write button could be used to both add and delete notes.

jmamma:
d62a7c2 <- I just doubled our efficiency.

jmamma:
Strange bug with del_note()

If I zoom in to finest resolution. And make two short notes. If I click in the centre of them, the note on the right is incorrectly  deleted.

```
add note
start_step = 2
end_step = 3
start_utiming = 8
end_utiming = 4
```

```
note_idx_on = 2
note_idx_off = 3
ev.micro_timing = 4
ev_j.micro_timing = 4
i = 2
j = 3
"deleteing" = deleteing
```

The microtiming of the note deleted is wrong...  (this is why it is deleted, the position of the note is incorrectly calculated and it passes the delete guard)

yatli:
something else that could be related, in the finest resolution, the trig interface doesn't map to the entire FOV.

jmamma:
> something else that could be related, in the finest resolution, the trig interface doesn't map to the entire FOV.

notes entered using the cursor... it's the finest resolution. you can see the microtiming in the debug.

jmamma:
```
i = 2
active_track.events[i].micro_timing = 8
i = 3
active_track.events[i].micro_timing = 4
```
After I store the note, microtiming is okay..

jmamma:
Problem is here in del()

ev_idx = 1
ev_idx_j = 3

^ ev_idx = 1  should be 2.

yatli:
I wrote a whole bunch of bugs

jmamma:
No more bugs please :sob:

yatli:
😅😅😅

jmamma:
Still one more issue... deleting note at finest resolution doesn't always remove the note_on event. 
*shrug*

jmamma:
```
add note
note_start = 2
note_end = 4
note_idx_on = 0
note_idx_off = 1
i = 0
j = 1
```

```
del note
note_start = 2
note_end = 4
i = 0
j = 1
"deleteing" = deleteing
note_idx_off = 1
note_idx_on = 0
```

```
537           DEBUG_DUMP("deleting");
538           active_track.remove_event(note_idx_off);
539           active_track.remove_event(note_idx_on);
```

I think there's a problem with remove_event still ... 

If I dump the event list after delete.. the note_on event is still there...

yatli:
on finest resolution, note start and end should fall into the same bucket (step). could be relevant.

jmamma:
note_start = 2
note_end = 4

in steps...

i also added 
129d3c7

yatli:
The add_event was bugged, last event not preserved.

yatli:
Does ext track save work for you? Notes do not persist on my side.

jmamma:
```
Does ext track save work for you? Notes do not persist on my side.
```

I haven't tested it yet...

jmamma:
add_note() seems more broken now.

yatli:
how? it appears to be ok for me.

jmamma:
Might not be as bad.. there are still some bugs.. at finest resolution.. but at low res too.. deleting notes doesn't appear to delete the note_on..

Getting late...

yatli:
one more quick question, about the repro note length? do I have to use the shortest possible notes?

jmamma:
no.. not necessarily.

yatli:
Ok I'll try to repro this one.

jmamma:
If i create a note.. then add a note behind it... the event ordering is messed up.


```
del note
Add note
start_step * timing_mid + start_utiming - timing_mid = 18
end_step * timing_mid + end_utiming - timing_mid = 22
start_step = 3
end_step = 4
i = 0
active_track.events[i].micro_timing = 6
i = 1
active_track.events[i].micro_timing = 4

del note
Add note
start_step * timing_mid + start_utiming - timing_mid = 13
end_step * timing_mid + end_utiming - timing_mid = 17
start_step = 2
end_step = 3
i = 0
active_track.events[i].micro_timing = 7
i = 1
active_track.events[i].micro_timing = 6
i = 2
active_track.events[i].micro_timing = 5
i = 3
active_track.events[i].micro_timing = 4

```

See that micro timing becomes interlaced!!

jmamma:
I suspect add_event is trying to be two things... insert_event and add_event.

jmamma:
Hmm. i think the interlacing might be okay.

Note 1 start step and note 2 end step share the same bucket..

jmamma:
~Okay. i think it's just a problem with the drawing algorithm.~

@yatli We will need to implement bucket re-ordering on add_event. 

Order:
1) CC
2) NoteOff
3) NoteOn


^

If note_on and note_off in same bucket... the note_offs must come first.

jmamma:
still broken...

jmamma:
🚀 🚀 🚀  🌔

jmamma:
@yatli 
a5c3e51 <- can you double check him.

yatli:
👍👍👍 I was thinking of something similar.

yatli:
@jmamma how about de2691c ?

Edit: Doesn't look good, notes messed up.

yatli:
Doesn't look good, notes messed up.

jmamma:
clean build?

yatli:
yes, clean build. removing a note ends up with another note showing up

only repros when removing the "only" note.

jmamma:
Are you recording from Chromatic page first?

jmamma:
It's working perfectly for me when editing fresh from the piano roll.

When recording from chromatic, wrap around notes can cause problems with delete.

yatli:
Ah, after a clear trk it's working now.
Must be related to save not working.

I was trying to edit the notes after chaining everything.

avr/cores/megacommand/MCL/ExtSeqTrack.cpp
```diff
@@ -60,30 +60,15 @@ uint16_t ExtSeqTrack::add_event(uint8_t step, ext_event_t *e) {
...
-          break;
-        }
-      }
-    }
-    else {
-    //If note on, insert at end
-    idx = end;
-    }
+  while(idx < end && events[idx].micro_timing < e->micro_timing) {
+    ++idx;
```
jmamma, line 64: @yatli we should still order based on event_type.

note_on event should come after note_off if microtiming matches...

yatli: roger.

jmamma:
Some bad news.

The piano roll rendering algorithm is unusable once we hit 128 step length, and 10-20 notes. 
FPS drops heavily.

jmamma:
I'll see if I can optimize it a bit. Otherwise we may have to pre-render.

yatli:
A4 keys aren't working in piano roll

jmamma:
> A4 keys aren't working in piano roll

You mean, cursor position? 

It's working for me. Check A4 MIDI config.

avr/cores/megacommand/MCL/SeqExtStepPage.cpp
```diff
@@ -589,7 +589,7 @@ bool SeqExtStepPage::handleEvent(gui_event_t *event) {
     uint8_t track = event->source - 128;
 
     if (device == DEVICE_A4) {
-      track -= 16;
+      return true;
```
jmamma, line 695: this broke your A4. @yatli

jmamma: ignore... UART2 used for MIDI notes ?

yatli: Nope, that was an effort trying to make it work.

yatli: I'm wondering why A4 note events don't show up in midi2 callbacks.
Keyboard CFG=Ext

jmamma: it's working for me...

jmamma: new firmware? keyboard needs to be set to transmit notes.

yatli: It works in the step edit page, triggers MD notes

yatli: eh, my bad. auto track channel was set to 9

jmamma: MIDI newbie :-p

yatli: the correct config should be, midi out ch -> TRK; was set to auto.

jmamma: That reminds me, we need channel selection in ext track menu.

jmamma:
```
 50 /// 24-bit ext track event descriptor
 51 struct ext_event_t {
 52   /// true for lock, false for Midi note
 53   bool is_lock : 1;
 54   /// effective when is_lock is true
 55   uint8_t lock_idx : 3;
 56   uint8_t cond_id : 4;
 57   /// for Midi note: note on/off
 58   /// for plock: lock engage/disengage
 59   bool event_on : 1;
 60   /// for Midi note: pitch value
 61   /// for plock & lock engage: lock value
 62   /// for plock & lock disengage: ignored
 63   uint8_t event_value : 7;
 64   /// micro timing value
 65   uint8_t micro_timing;
 66 };
 67 
```

have we any room for slide ?

yatli:
> have we any room for slide ?

I left that as a TODO:

Given that the space is so tight, packing another bit is not viable.
However, we can use a modifier event to indicate that the next N events are slide-enabled.

yatli:
@jmamma as simple as that 89dc6f1

yatli:
perf boost ready.

yatli:
Out for the next trip tmr.
I really prefer chunky trip than "chatty" but there's a lonely furball at home, waiting to be fed/pet.

jmamma:
> perf boost ready.

Great work on the optimizations. Night and day difference.

jmamma:
5ede940 is causing lockup if change octaves during record... no idea why. 

🛏️ review in morning.

jmamma:
There is a bit of a problem when recording from Chromatic then trying to edit the resulting sequence.
Basically, the live record was less critical about overlapping notes of the same pitch.
 I've put in place a few mitigations to help cleanup the recording, by recycling some of the piano roll code:

- Notes are only recorded once note_off of same pitch is received.
- A 16 note buffer is used for this, per sequencer track. Max polyphony on record is therefor 16 notes.
- RAM usage is 16 * 3 = 48 bytes per ext track.
- Notes are now deleted before they are recorded.
- Note deletion is now recursive, until no more notes can be deleted under cursor/note width.

jmamma:
PianoRoll Todo: 

- [x] Add record modes to SeqStep and SeqExtStep pages. (Possibly consolidate behaviour with SeqPtcPage, already started looking at this could be messy).
- [ ] Parameter locks for ExtSeqTrack, with slides.
- [ ] Parameter lock editor for ExtSeqTrack, with slide interpolation 

Bugs:

- [x] Live record can lead to overlapping notes of same pitch. I can't replicate this problem when entering notes on the piano roll,. My mitigation strategy was to making recording notes identical to behaviour of inserting notes on the piano roll, by deleting any overlapping notes first. This however does not seem to work in practice, and I can't workout why....

yatli:
Finally back home. I just want to go to work. No more vacations 🙃

yatli (commented):
👍

jmamma:
RE Locks: 
1) We don't necessarily need a slide bit. We could just have all lock data interpolated for ext tracks.

2) A4 uses 14bit CC messages for some parameters, like Filter. I wonder if we can support this by using adjacent events in the event array.

3) Need to create a mapping from CC->A4Sound and vice versa.

yatli:
> all lock data interpolated for ext tracks

understood. In this case, if we want a lock for 5 steps and then another lock, we do:

```
step1: lock value1
...
step5: lock value1
step5: lock value2
```

correct?

avr/cores/megacommand/MCL/SeqPage.cpp
```diff
@@ -617,7 +617,7 @@ void SeqPage::draw_mask(uint8_t offset, uint8_t device,
     draw_mask(offset, mask, active_track.step_count, active_track.length,
               oneshot_mask, slide_mask, show_current_step);
 
-    if (recording);
```
yatli: 😱

jmamma:
> > all lock data interpolated for ext tracks
> 
> understood. In this case, if we want a lock for 5 steps and then another lock, we do:
> 
> ```
> step1: lock value1
> ...
> step5: lock value1
> step5: lock value2
> ```
> 
> correct?

yes.

yatli:
> A4 uses 14bit CC messages for some parameters

if we allocate another byte for `ext_event_t`, we get high resolution ext locks, plus velocity for notes without locks.
~~currently BANK1_FILE_ENTRIES_END == 54998, cost for extra byte in ext_event_t == 6 * 512~~ edit: stale numbers, correction:

yatli:
```c++
// as of commit  33e243afc758081dc6eb244e42ae61e1e0de09c0
// the track sizes are:
// MDTrack 534
// A4Track 1957
// DeviceTrack 7
// EmptyTrack 1957
// FXTrack (TODO) 43


// 16x MD tracks
// GRID1 tracks start at 0x6B60
constexpr size_t BANK1_MD_TRACKS_START = BANK1_SYSEX2_DATA_START + SYSEX2_DATA_LEN;
// AUX tracks start at 0x8CC0
constexpr size_t BANK1_AUX_TRACKS_START = BANK1_MD_TRACKS_START + GRID1_TRACK_LEN * NUM_MD_TRACKS;
// 6x A4 tracks
// GRID2 tracks start at 0x8D16
constexpr size_t BANK1_A4_TRACKS_START = BANK1_AUX_TRACKS_START + FX_TRACK_LEN * NUM_FX_TRACKS;
 
// 256x file entries (16 bytes each)
// Start at 0xBAF4
constexpr size_t BANK1_FILE_ENTRIES_START = (BANK1_A4_TRACKS_START + GRID2_TRACK_LEN * NUM_A4_TRACKS);
constexpr size_t BANK1_FILE_ENTRIES_END = (BANK1_FILE_ENTRIES_START + 16 * NUM_FILE_ENTRIES);

// At 0xCAF4
```

yatli:
before the extra byte:
13580 bytes left 😎

after:
end at 55142, 10394 bytes left.

jmamma:
That’s a good way to do it. But I’m having second thoughts about explicitly supporting hi-res parameters. 

Technically in record mode hi res CC would work without any special implementation as the extra bits would just occupy an additional  pararm_idx

yatli:
FYI a2f33b1c215be4d866a9be51d7c91f24a321ad97

> Technically in record mode hi res CC would work without any special implementation as the extra bits would just occupy an additional pararm_idx

You mean, on the MIDI side it's actually two different CC targets?

jmamma:
> FYI [a2f33b1](https://github.com/jmamma/MIDICtrl20_MegaCommand/commit/a2f33b1c215be4d866a9be51d7c91f24a321ad97)
> 
> > Technically in record mode hi res CC would work without any special implementation as the extra bits would just occupy an additional pararm_idx
> 
> You mean, on the MIDI side it's actually two different CC targets?

Yes. Look at the A4 manual

yatli:
A4 newbie 🙄

yatli:
That's interesting actually. It must then implement a state machine so that the hires knob is not updated with just one CC event

jmamma:
> A4 newbie 🙄

I discovered this today also 😀. Had to observe MIDI output to be sure

yatli:
> all lock data interpolated for ext tracks
> 
> understood. In this case, if we want a lock for 5 steps and then another lock, we do:
> 
> step1: lock value1
> ...
> step5: lock value1
> step5: lock value2
> correct?
> 
> yes.

plus we have a lock engage/disengage bit to determine the continuity of slide:

image

lock engage = keep sliding (key frames marked with 3x3 squares)
lock disengage = back off to original value, break the slide (red dotted line).

An Elektron-style lock is an engage-disengage pair.
A continuous envelope is a engage-engage-...-disengage chain.

jmamma:
So you want to go with the extra bit byte for velocity. +1 bit for lock_engage. ?

yatli:
The 1 bit lock_engage is already here :)

yatli:
for velocity and hi-res (not A4, but general hi-res msg) we can implement the state machine. no need to extend the event structure as it would be a total waste for other events.

jmamma:
> for velocity and hi-res (not A4, but general hi-res msg) we can implement the state machine. no need to extend the event structure as it would be a total waste for other events.

I don't follow. Are you intending to use the additional byte in the event structure?

yatli:
No. I want to use the current event structure.
Velocity is just a special lock target that doesn't allow sliding.
A4 hires events are just paired CC events.

For "real" hi-res events, we cannot guarantee it's just two bytes (could be more), so we set up state machines for them.

jmamma:
It seems we have a bit of a dilemma.

Separate byte of hi-res events is a waste of a byte when only low-res is used. 
A separate event for velocity is a waste of state/timing/condition bits.

jmamma:
I'm thinking we have two event arrays. One for note with velocity, the other for locks (7bit)?

yatli:
Is note off velocity really useful?

jmamma:
Ah good point. Not useful at all.

jmamma:
Clarification: Not useful to us. Some synths might use it for controlling decay. but i've never had a practical use for it.

yatli:
I'm having trouble thinking of what a "note off velocity" even means :D

Like, releasing a key on a piano... slow or fast?

jmamma:
```On a real piano, if you lift the key very slowly, the damper will fall to the string slowly, and it does create a different sound compared to a fast release. It's subtle, though... and not something I'm sure anyone would hear in very much normal actual playing.```

**Comment by yatli** [comment]
Oh ok, I remember hearing something like that in a record. Some rich harmonics.

**Comment by jmamma** [comment]
Do we steal a param_idx for velocity? We would also need to store a reference velocity somewhere.

**Comment by yatli** [comment]
Yeah, I prefer doing it this way, allocating a special param target.
And then we store the last known velocity somewhere for a track for new notes.

**Comment by jmamma** [comment]
Need to work out where to put 

locks_params_orig[NUM_LOCKS] (the reference / last known parameter values.)

On MDSeqTrack this is ephemeral, and is populated on track load from MDKit.params

--

Thinking about making it persistent for ExtSeqTrack, as general MIDI device won't be able to populate this data on load.

**Comment by yatli** [comment]
Our chief is back!

**Comment by jmamma** [comment]
Treating velocity as a separate lock param might be too convoluted.

1) For multiple notes on the same step, you would need a separate event for each to represent velocity, These would need to be ordered, and cleaned up correctly upon note deletion.
2) 1 could be mitigated by limiting velocity to a single value for all notes on a single step. 
3) If you go with 2, you could just implement a static array of 128 bytes, representing velocity values for each step.

Reference velocity wont save much space when recording from a velocity sensitive keyboard, as values will be inconsistent.

**Comment by jmamma** [comment]
@yatli thoughts?

**Comment by jmamma** [comment]
Current Event Schema (No Velocity)
512 * 3  =  1536

Current Event Schema (Velocity is treated as lock param: additional complexity particularly if aligning velocity with specific note event)
512 * 3  =  1536

Add additional byte to Event for Velocity. (Each event gets on/off velocity. additional byte is wasted with locks)
512 * 4 = 2048

Add 128 byte array for velocity (Events on same step share same velocity)
512 * 3 + 128 = 1664

**Comment by jmamma** [comment]
velocity is working.

**Comment by jmamma** [comment]
Slides are done, need to consolidate the code with MDSeqTrack.cpp

**Comment by jmamma** [comment]
We've exhausted the progmem again...

**Comment by jmamma** [comment]
Tried to include another font, and/or extend  
```16 const menu_option_t MENU_OPTIONS[] PROGMEM = {```
results in corrupt binary.

**Comment by yatli** [comment]
a little bit busy with day job these days.. probably got time to drop in next week..

**Comment by jmamma** [comment]
all good. i should be able to finish up the pianoroll editor. 

i've hit an issue with program size and corrupt binary... still need to do a few rounds of compression in this branch. but i suspect it's something we'll run into again soon.

**Comment by jmamma** [comment]
~progmem corruption was caused by faulty arduino mega.~

```
avrdude: verification error, first mismatch at byte 0x3c000
         0x00 != 0x2f
avrdude: verification error; content mismatch
```

The flash mem was only good up to 245760 bytes.

**Comment by jmamma** [comment]
Happening on another board.

ArduinoMega bootloader is 8KB. D'oh.

Edit: looks like a bug in the bootloader, not initialising all of the flash memory?

https://github.com/MarlinFirmware/Marlin/issues/17598

**Comment by jmamma** [comment]
Hmm.. I just spent a day debugging a GUI menu routine to update the CC parameter of a lock.

```
SeqExtStepPage::loop() {
auto active_track = mcl_seq.ext_tracks[last_ext_track];
active_track.locks_params[c] = blah
 }
```
The memory assignment above would be reset, after existing loop()

Removing use of auto solved the problem 😱

**Comment by yatli** [comment]
> Removing use of auto solved the problem scream

should be `auto &active_track`?

**Comment by jmamma** [comment]
auto active_track = mcl_seq.ext_tracks[last_ext_track];

is this just a copy then ?

File: avr/cores/megacommand/MCL/ExtSeqTrack.h [hunk 5]
```diff
@@ -70,6 +70,13 @@ struct ext_event_t {
     if (this->micro_timing != that.micro_timing) {
       return this->micro_timing < that.micro_timing;
     }
+    /*
+    // order by lock
```
jmamma, line 74: @yatli can you fix this. we want locks before notes.

yatli:
> is this just a copy then ?

yes...

avr/cores/megacommand/MCL/ExtSeqTrack.h
yatli: will do.

jmamma:
- [x] ExtSeqTrack Slides are still broken... probably something in the find_next_locks code
- [x] Edge cases for drawing locks need work. When start / end of a slide is outside the fov, we need to correctly calculate the y position at x = 0 / x = fov_end, using the gradient of the line. 
- [x] Sometimes no lock data will be drawn, then after adding a number of additional events it renders... not sure why.

jmamma:
Drawing recorded locks appears to be quite broken 👎

jmamma:
@yatli this one is done. 😓 

only took 5 months 🥇

PR #126, opened 2020-08-03
Monomachine

yatli:


jmamma:
Cool.

To be a pain. For MidiCtrl consistency. Move from Struct -> Class ? 
Driver connect/disconnect could then override virtual method from a parent Driver.

yatli:
How do two devices share probing code?

jmamma:
> How do two devices share probing code?

inheritance ?

yatli:
Hmm yeah I mean, how are two devices supposed to share anything? 🤔
Most shared logic are related to GUI and device info update, and already extracted in `probePort`. The drivers themselves are highly device-specific.

jmamma:
> Hmm yeah I mean, how are two devices supposed to share anything? 🤔
> Most shared logic are related to GUI and device info update, and already extracted in `probePort`. The drivers themselves are highly device-specific.

Yeah I see your point.

yatli:
``` c++
        MidiIDSysexListener.setup(pmidi_class);
        MidiUart.set_speed((uint32_t)31250, port);
#ifdef OLED_DISPLAY
        auto oldfont = oled_display.getFont();
        prepare_display();
        if (drivers[i].icon) {
          oled_display.drawBitmap(14, 8, drivers[i].icon, 34, 42, WHITE);
        }
        oled_display.display();
#else
        GUI.clearLines();
        GUI.setLine(GUI.LINE1);
        GUI.put_string_at_fill(0, "Peering...");
        LCD.goLine(0);
        LCD.puts(GUI.lines[0].data);
        LCD.goLine(1);
        LCD.puts(GUI.lines[1].data);
#endif


```

Can these things be pulled to the outer loop so that it's done only once for a driver?

Edit: I think so. It's not included in the original retry loops anyway.

jmamma:
yeah i think it's okay.

also, secretly i've been dropping support for the hd44780. i think it's time.

yatli:
That'll be a relief :)

yatli:
I'm trying to bring up basic mnm peering but it doesn't work..
Need a bit help @jmamma

jmamma:
MNM.getCurrentKit(CALLBACK_TIMEOUT) not working?

jmamma:
You need to enable the sysex listener for MNM

void MCL::setup() {
 MNMSysexListener.setup(&Midi2);

--

Also, we need to increase sysex listener buffer because it's currently maxed out

```
$ grep -r NUM_SYSEX_SLAVES .
./MidiSysex.cpp:  for (int i = 0; i < NUM_SYSEX_SLAVES; i++) {
./MidiSysex.hh:#define NUM_SYSEX_SLAVES 4
```

jmamma:
We should probably have a menu/cfg item allowing you to specify which devices to listen for on which ports.

yatli:
Yeah. I've left the `port` parameter in the drivers ;)

yatli:
> Also, we need to increase sysex listener buffer because it's currently maxed out

It's maxed out on port1, should be fine for port2?

jmamma:
ah you're right. it should be a separate ~buffer~ array* on 2.

yatli:
I suspect `MNMSysex.cpp` doesn't work at all(

yatli:
Fixed it by copying the logic from `MDSysex` to `MNMSysex`.
Turbo works fine at 8x.

yatli:
I think I'll just stop here before this thread becomes MNM dev thread. After this gets in I'll work on the MNM part.

yatli:
btw for some reason MNM doesn't support Midi Id either.

0.0

jmamma:
> Fixed it by copying the logic from `MDSysex` to `MNMSysex`.
> Turbo works fine at 8x.

I suspected the Sysex side needed updating.
There were a whole lot of untested changes after moving the sysex buffers to memory bank 1

jmamma:
> I think I'll just stop here before this thread becomes MNM dev thread. After this gets in I'll work on the MNM part.

you're doing good work here. cheers.

jmamma:
> btw for some reason MNM doesn't support Midi Id either.
> 
> 0.0

sigh.

yatli:
image

Does this feel right? :p

yatli:
The MnM active sense behavior is weird.
The negotiation is good, and it connects with speed up to 8x (ok)
After handshake, it sends back active sense ack every 300ms (ok)
Until I press a trig -- the next active sense ack is delayed (not ok, the task will see MnM as timeout)

The workaround: enable midi clock out on MnM -- that will act as active sense tick.

I guess the logic is that a normal midi message also counts as active sense tick, so when I press a trig and send out the message, it skips one tick...

Too lazy :/

yatli:
@jmamma something weird is happening.

```c++
bool MNMGlobal::fromSysex(MidiClass *midi) {
  uint16_t offset = 5;
  uint16_t len = midi->midiSysex.recordLen - 5;

  //DEBUG_PRINT_FN();
  DIAG_DUMP(offset);
  DIAG_DUMP(len);
```

Upon boot, `len` gets wrong a few times. Then return to correct values.

avr/cores/megacommand/A4/A4Sysex.cpp
```diff
@@ -15,7 +15,7 @@ void A4SysexListenerClass::start() {
 
 void A4SysexListenerClass::handleByte(uint8_t byte) {
 
-  if (isA4Message && MidiSysex2.len == sizeof(a4_sysex_hdr)) {
+  if (isA4Message && MidiSysex2.recordLen == sizeof(a4_sysex_hdr)) {
```
jmamma, line 18: these changes will also need to be made in the MD library

avr/cores/megacommand/Midi/MidiSysex.h
```diff
@@ -101,54 +101,23 @@ class MidiSysexClass {
   }
 
   uint8_t getByte(uint16_t n) {
-    if (n < maxRecordLen) {
+    if (n < sysex_bufsize) {
       // Retrieve data from specified memory buffer
-      if (recordBuf != NULL) {
```
jmamma, line 106: okay, we probably dont need to support array buffer.

avr/cores/megacommand/A4/A4Sysex.cpp
yatli: the `handleByte` of listeners are never called?

yatli:
Super broken, most sysex functionalities don't work. MD needs reset.

avr/cores/megacommand/A4/A4Sysex.cpp
jmamma: ignore. i thought len -> recordLen was variable name replacement.

---

handleByte wont be called on receive. we only process sysex messages when they are complete. by calling either end(); or end_immediate();

end_immediate() occurs during ISR.
end() occurs during main loop.

jmamma:
How much memory do you want to keep for uart2 sysex buffer?

Do we want to investigate a shared buffer implementation?

yatli:
Going to need your help to debug this one. I have no clue why things are breaking down now...

yatli:
```
func_call: void MCLActions::store_tracks_in_mem(int, int, uint8_t)
fetching pattern
kit wrong length
len = 5403
could not receive kit
```

yatli:
@jmamma 80ad756 fixes a bad bug caused by the fw running faster than before.

Some notes:

```c
  if (msgType == 0x72) {
    // XXX wrong reply. fix the firmware!
    if (MDSysexListener.sysex->getByte(begin) == 0x30) {
      ((uint8_t *)&(fw_caps))[0] = MDSysexListener.sysex->getByte(begin+1);
      ((uint8_t *)&(fw_caps))[1] = MDSysexListener.sysex->getByte(begin+2);
    }
```

The FW doesn't reply with a 0x30 header... Just caps. This will never execute.


```c++
void MDClass::activate_trig_interface() {
  uint8_t data[3] = {0x70, 0x31, 0x01};
  sendRequest(data, sizeof(data));
  waitBlocking(); // <------ this
}
```

All these are lacking the `waitBlocking`, meaning the FW will get back with a wrong message that disrupts a few things.
Before this branch the sysex handler wasn't running as fast as now, so when the 0x72 message comes back, no new listener callback is registered, and we simply ignored that.

TODO: we can also clear `callSysexcallbacks` when we register a listener.


image

^^^ I really don't know about this one. Fixes wrong recordLen.

avr/cores/megacommand/MD/MD.cpp
```diff
@@ -187,21 +190,25 @@ bool MDClass::get_fw_caps() {
 void MDClass::activate_trig_interface() {
   uint8_t data[3] = {0x70, 0x31, 0x01};
   sendRequest(data, sizeof(data));
+  waitBlocking();
```
jmamma: These are asynchronous requests. why do we need to wait ?

avr/cores/megacommand/Midi/MidiSysex.h
```diff
@@ -113,7 +113,8 @@ class MidiSysexClass {
     if (recordLen < sysex_bufsize) {
       // Record data to specified memory buffer
       // Write to sysex buffers in HIGH membank
-      putByte(recordLen++, c);
+      putByte(recordLen, c);
+      ++recordLen;
```
jmamma, line 119: :confused:

avr/cores/megacommand/MD/MD.cpp
yatli:
```
send gateway command
send getBlocking command
waiting for callback to trigger
gateway command replies
callback accepted the sysex command
read the wrong message
getBlocking command replies, and ignored
```

jmamma: I don't follow,

activate_trig_interface() just sends a sysex message to the MD.

There should be no waiting after this ?

yatli: > activate_trig_interface() just sends a sysex message to the MD.

Each 0x72 gateway message has a reply.

jmamma: ah i see that now.

jmamma:
```
Before this branch the sysex handler wasn't running as fast as now, so when the 0x72 message comes back, no new listener callback is registered, and we simply ignored that.
```
Can you elaborate on this? Which 0x72 message? and how is the callback not registered before the request?

yatli:
We do a lot of 0x72 before getBlockingKit(), for example, during midi active peering we do `get_fw_caps`, and during GridSave, we `deactivate_trig_interface`.

If the 0x72 ack reply comes back earlier than the getBlockingKit() call, it will be safely ignored.

jmamma:
```   
 // XXX another 0x72 message is received after get_fw_caps, and getBlockingKit accepts the wrong message
```

Are you saying     if (!MD.get_fw_caps()) {
generates 2,  0x72 reply messages?

yatli:
Yes.. No idea why.

jmamma:
We need to look at the MIDI exchange.
verify that MCL is only sending one request. 
and/or verify that MD is sending only one reply

yatli:
The two messages dumped from end_immediate (isr):

```
getByte(5) = 114 (0x72)
getByte(6) = 34 (0x22)
getByte(7) = 2 (0x2)
getByte(8) = 0 (0x0)

getByte(5) = 114 (0x72)
getByte(6) = 30 (0x1E)
getByte(7) = 0 (0x0)
getByte(8) = 0 (0x0)
```

avr/cores/megacommand/MD/MD.cpp
jmamma: > The two messages dumped from end_immediate (isr):
> 
> ```
> getByte(5) = 114 (0x72)
> getByte(6) = 34 (0x22)
> getByte(7) = 2 (0x2)
> getByte(8) = 0 (0x0)
> 
> getByte(5) = 114 (0x72)
> getByte(6) = 30 (0x1E)
> getByte(7) = 0 (0x0)
> getByte(8) = 0 (0x0)
> ```

Is this undefined behaviour? Looking at the MD side we have a switch statement, but there's no default case for when the message type is unrecognized.

yatli:
Doesn't make sense to me. 
That 0x72 0x1E message seems OK (0b00011110), fw cap reply, but the first one? Looks like a track index query response.

Edit: there's indeed a track index query before get_fw_caps.

yatli:
2e61e7e

just a stupid error causing the blocking callbacks to arbitrarily unblock...

jmamma:
```
The two messages dumped from end_immediate (isr):

getByte(5) = 114 (0x72)
getByte(6) = 34 (0x22)
getByte(7) = 2 (0x2)
getByte(8) = 0 (0x0)

getByte(5) = 114 (0x72)
getByte(6) = 30 (0x1E)
getByte(7) = 0 (0x0)
getByte(8) = 0 (0x0)
```

Is this undefined behaviour? Looking at the MD side we have a switch statement, but there's no default case for when the message type is unrecognized.

yatli:
> Is this undefined behaviour? Looking at the MD side we have a switch statement, but there's no default case for when the message type is unrecognized.

Not a problem. No default case = empty default case, so that it will reply with an empty message (0x72 0xF7)

jmamma:
> [2e61e7e](https://github.com/jmamma/MIDICtrl20_MegaCommand/commit/2e61e7eac80496b9cdc795831c5e8b65a019ad69)
> 
> just a stupid error causing the blocking callbacks to arbitrarily unblock...

is everything working now? or still problems?

yatli:
yeah it works now. :D

yatli:
This is a bit daunting -- cannot get global send to work, though I'm almost sure my data codec implementation is good.

jmamma:
What's not working exactly?

Do you get an SYSEX error on the MNM?

yatli:
No error, it just ignores the global data.
I even tried to capture a global dump from MNM and replay, and send it back -- didn't work either

jmamma:
what if you force receive from the sysex menu?

You should be able to use "verify" to validate the data as well.

yatli:
Good idea.. Let me try it.

yatli:
It works from the sysex recv menu... T.T
I wasted some 6 hours on this

jmamma:
submit a bug report :D

yatli:
This will be fixed from MMX firmware.

yatli:
@jmamma I'm excited to bring in b923b20 -- unifies a lot of things!

jmamma:
Nice job. It never occurred to me to consolidate this code, should save some bytes!
Wonder how many elektron machines we can squeeze in concurrently.

yatli:
This will be helpful for refactoring MCLActions as well. We now have generic MidiDevice and ElektronDevice so that we can do:

```
MidiDevice* dev = midi_active_peering.get_device(UART_PORT2);
DeviceTrack* ptrack = empty_track.load_from_grid<ExtTrack>(n, row);
if (dev->connected && dev->compatible(ptrack)) { // dev takes care of track compatibility!
  sound_send[grid_col] = 1;
}
```

yatli:
> We probably don't need 8x callback vectors.

Maybe 1 is enough! (diff = ~630bytes)

jmamma:
RAM diff, or pgmem diff ?

also. this commit is enormous.

yatli:
That's RAM diff. progmem-wise, shrinking it to 1 will not help that much.

jmamma:
> 
> 
> That's RAM diff. progmem-wise, shrinking it to 1 will not help that much.

just checking.

jmamma:
> 
> 
> This will be helpful for refactoring MCLActions as well. We now have generic MidiDevice and ElektronDevice so that we can do:
> 
> ```
> MidiDevice* dev = midi_active_peering.get_device(UART_PORT2);
> DeviceTrack* ptrack = empty_track.load_from_grid<ExtTrack>(n, row);
> if (dev->connected && dev->compatible(ptrack)) { // dev takes care of track compatibility!
>   sound_send[grid_col] = 1;
> }
> ```

excellent.

yatli:
The driver framework is also updated. Now the device classes themselves are the drivers, the descriptors, and generic devices:

```
static MidiDevice* port1_drivers[] = {
  &MD
};

static MidiDevice* port2_drivers[] = {
  &MNM,
  &Analog4,
  &generic_midi_device,
};
```

TODO: automatically plug in NullDevice (yeah, /dev/null) if nothing is online.

jmamma:
> 
> 
> The driver framework is also updated. Now the device classes themselves are the drivers, the descriptors, and generic devices:
> 
> ```
> static MidiDevice* port1_drivers[] = {
>   &MD
> };
> 
> static MidiDevice* port2_drivers[] = {
>   &MNM,
>   &Analog4,
>   &generic_midi_device,
> };
> ```
> 
> TODO: automatically plug in NullDevice (yeah, /dev/null) if nothing is online.

yep I saw this. pretty neat stuff.

yatli:
> Wonder how many elektron machines we can squeeze in concurrently.

How about:

`class VstDevice: public MidiDevice`

And then turn MegaCom into a VST host(

jmamma:
> 
> 
> > Wonder how many elektron machines we can squeeze in concurrently.
> 
> How about:
> 
> `class VstDevice: public MidiDevice`
> 
> And then turn MegaCom into a VST host(

you want to run DSP @ 16MHz? 😉

yatli:
Good enough for SIDStation-ish sound, I guess?

jmamma:
better to create a voice board to connect via expansion port.
could be analog or dsp based. MCL is :brain:

yatli:
Yeah that's it. MegaCom for the transport layer.

yatli:
@jmamma 80203ba "demos" some changes in MCLActions

yatli:
image

... Welcome to MegaCommand Live

(Did it just say "crap"?)

jmamma:
```
$ grep -r crap .
./Midi/MidiSDS.cpp:      DEBUG_PRINTLN("crap");
```
Yep looks like something i'd write.

jmamma:
```
$ grep -r crap .
./Midi/MidiSDS.cpp:      DEBUG_PRINTLN("crap");
```
Yep looks like something i'd write.

yatli:
Hah I thought it was self-motivated!

yatli:
```
event->source = 129
129
note event
2
0
event->source = 128
128
note event
2
0
0
note off
1
note off
event->source = 128
128
note event
0
2
notes all off
deactiviating trig interface
func_call: void MCLActions::write_tracks(int, int)
kit okay
func_call: void MCLActions::send_tracks_to_devices()
func_call: DeviceTrack* DeviceTrack::init_track_type(uint8_t)
534
func_call: DeviceTrack* DeviceTrack::init_track_type(uint8_t)
i = 0
func_call: DeviceTrack* DeviceTrack::init_track_type(uint8_t)
534
func_call: DeviceTrack* DeviceTrack::init_track_type(uint8_t)
i = 1
latency
56
BEFORE GUI LOOP
event->source = 129
129
note event
0
2
notes all off
func_call: void MCLActions::write_tracks(int, int)
```

any idea why write_tracks -> GUI.loop -> write_track forms a deadloop?

jmamma:
what's the context?

jmamma:
are you automating this ?

yatli:
Press write, then two trigs.
It calls into write page :: handleEvent -> MCLActions::write_tracks -> ... -> GUI.loop() -> pumps out note off events again -> handleEvent -> write_tracks -> .... stack overflow?

jmamma:
catching up on your changes...

yatli:
let me push it now.

jmamma:
kk. it sounds like you're getting some undefined behaviour. possibly due to memory overflow/corruption.

yatli:
I think I figured it out..

If I release the two trigs one after another it will not crash.
If I release the two trigs simultaneously it will.

The `notes_all_off` predicate turns true before the second event arrives at handleEvent, because note_off_event is triggered via ISR.
So the first note off event will trigger write track, with the second pending.

jmamma:
sequencer running or stopped?
this will determine send_tracks_to_devices vs prepare_next_chain

yatli:
Sequencer stopped.

jmamma:
how does this trigger a crash?

the chain page should close after detecting all_notes_off.

yatli:
> the chain page should close after detecting all_notes_off.

It only does it after mcl_actions.write_tracks finish, which includes another GUI loop driving loop.

jmamma:
```
24 bool GridWritePage::handleEvent(gui_event_t *event) {
125   // Call parent GUI handler first.
126   if (GridIOPage::handleEvent(event)) {
127     return true;
128   }
129   DEBUG_DUMP(event->source);
130   if (note_interface.is_event(event)) {
131     DEBUG_PRINTLN("note event");
132     trig_interface.send_md_leds();
133     if (note_interface.notes_all_off()) {
134       DEBUG_PRINTLN("notes all off");
135       if (BUTTON_DOWN(Buttons.BUTTON2)) {
136         return true;
137       } else {
138 #ifdef OLED_DISPLAY
139         oled_display.textbox("CHAIN SLOTS", "");
140         oled_display.display();
141 #endif
142         trig_interface.off();
143         mcl_actions.write_tracks(0, grid_page.encoders[1]->getValue());
144       }
145       GUI.setPage(&grid_page);
146       curpage = 0;
147     }
```

jmamma:
Ah crap!

370   while (clock_diff(myclock, slowclock) < md_latency_ms) {
371     GUI.loop();
372   };

^ probably him

yatli:
Yes you got it.

yatli:
One of the more sophisticated bugs...

jmamma:
I don't thinnk i've run in to this issue when testing. I must release the keys in a different way..

yatli:
To repro you'll have to release two trigs with very little time diff, so that the second event arrives before the first is processed.

jmamma:
So the reason that gui.loop /delay is there is because starting the sequencer during** kit transmission would cause corruption on the MD side.

So i added delay, such that the kit would finish transmitting before the internal sequencer is unmuted...

yatli:
Take a look at https://github.com/jmamma/MIDICtrl20_MegaCommand/pull/126/commits/afc205db0fe93e68d6dbc3b3287ae149a1c023d6#diff-6427ee8ba305bdf878dad651995a6fcd

It generalizes a lot of things.

jmamma:
> Take a look at [afc205d#diff-6427ee8ba305bdf878dad651995a6fcd](https://github.com/jmamma/MIDICtrl20_MegaCommand/commit/afc205db0fe93e68d6dbc3b3287ae149a1c023d6#diff-6427ee8ba305bdf878dad651995a6fcd)
> 
> It generalizes a lot of things.

Yeah, so we have a repeating pattern now. And you've suggested use func pointer to execute behaviour for each mode.

yatli:
See invocations of `sendKitParams` for how to dynamic dispatch the behavior

All device-specific code can be moved into the driver.

jmamma:
what do you mean by 

// No STL, no closure, no std::function, cannot make this generic...

yatli:
Ah that. Arduino doesn't have `std::function<T>`, otherwise I can further compress the bits

jmamma:
oh right stdlib. 

--
Do you need to push ? sendKitParams doesn't seem generic?

yatli:
The behavior is bound to the devices.
For example, see MD::sendKitParams or A4::sendKitParams -- they implement this differently.

yatli:
So in MCLAction there are only two concepts we need to care about:

- The device
- The device track

jmamma:
> The behavior is bound to the devices.
> For example, see MD::sendKitParams or A4::sendKitParams -- they implement this differently.

Yes, but I dont see the connection with respect to https://github.com/jmamma/MIDICtrl20_MegaCommand/commit/afc205db0fe93e68d6dbc3b3287ae149a1c023d6#diff-6427ee8ba305bdf878dad651995a6fcd

other than use of function pointer

yatli:
Ah ok. It's not in that specific patch.

But the patch does show evidences of generic processing: `MD.kit.toSysex(), a4_track->sound->toSysex()` are gone.

jmamma:
i like where this is all heading. keep going :rocket:

jmamma:
```      auto *pmem_track = empty_track2->load_from_mem(devs[grid]->track_type, grid_col);```

problem is grid2 has 2 devices... what's the plan?

**Comment by yatli** [comment]
Logically I see grid2 is actually two grids, one for sound one for FX

So in each driver, we further define the type of FX track it supports so as to guide MCLActions (we don't want to send incompatible FX events to a device anyway)

For MD it's the global FX
For A4 it's the CV/FX tracks
For MNM... I don't know. Mono/Poly switch?

**Comment by jmamma** [comment]
Okay, currently the grid variable is used in this way proj.select_grid(grid) which will direct reads/writes to the correct grid_file accordingly.

just need to make sure we dont do proj.select_grid(grid = 3);

**Comment by yatli** [comment]
> just need to make sure we dont do proj.select_grid(grid = 3);

haha no need. We can hard code this rule in MCLActions that everything after NUM_EXT_TRACKS in grid 2 is FX.

The logic:

- load up a fx track as DeviceTrack
- for device in port_devices, find device->fx_track_type == track->active
- type matches.

**Comment by yatli** [comment]
86a927f `#FAIL`

**Comment by yatli** [comment]
Can't get release build to work...

**Comment by jmamma** [comment]
problem?

**Comment by yatli** [comment]
panics on chain write.

**Comment by yatli** [comment]
could it be `GUI.loop()` triggering `GridTask::run` then re-entering `cache_next_tracks`?

**Comment by jmamma** [comment]
you can just comment that line out

**Comment by yatli** [comment]
testing:

```
  // note, do not re-enter grid_task -- stackoverflow
  GUI.removeTask(&grid_task);
  while (clock_diff(myclock, slowclock) < latency_ms) {
    GUI.loop();
  }
  GUI.addTask(&grid_task);
```

Edit: still crashes.

**Comment by yatli** [comment]
Commenting out GUI.loop: still crashes

**Comment by jmamma** [comment]
a lot of changes... are you writing a single track or everything?

**Comment by yatli** [comment]
just one track.

**Comment by yatli** [comment]
slowly stepping through the code with toggleLed2()....

**Comment by jmamma** [comment]
i usually just comment out large chunks until it starts working.

**Comment by yatli** [comment]
woah it starts to send pixels to the oled this time

**Comment by yatli** [comment]
"out... I want out..."

**Comment by yatli** [comment]
okay, it only happens when mnm is connected.

**Comment by yatli** [comment]
narrowed it down to this region:

```
  for (uint8_t i = 0; i < NUM_SLOTS; i++) {

    uint8_t grid_col = i;
    uint8_t grid = 0;
    // GRID 1
    if (i < GRID_WIDTH) {
      grid = 0;
    }
    // GRID 2
    else {
      grid = 1;
      grid_col -= GRID_WIDTH;
    }

    if (grid_col < devs[grid]->track_count) {
      SeqTrack *seq_track = mcl_seq.seq_tracks[i];
      mute_states[i] = seq_track->mute_state;
      seq_track->mute_state = SEQ_MUTE_ON;
    }

    proj.select_grid(grid);
    if (note_interface.notes[i] <= 1) {
      continue;
    }


    grid_page.active_slots[i] = grid_page.getRow();


    if (first_note == 255) {
      first_note = i;
    }

    auto *ptrack = empty_track.load_from_grid(grid_col, grid_page.getRow());
    if (ptrack->is_active()) {
      DEBUG_DUMP(i);
      ptrack->chain.store_in_mem(i, &(chains[0]));
      ptrack->load_immediate(grid_col);
      if (grid_col < devs[grid]->track_count) {
        send_masks[i] = 1;
      }
    }
  }
```

**Comment by jmamma** [comment]
SeqTrack *seq_track = mcl_seq.seq_tracks[i];
this will be out of bounds...

 SeqTrack *seq_tracks[NUM_MD_TRACKS + NUM_EXT_TRACKS];

NUM_MD_TRACKS + NUM_EXT_TRACKS < NUM_SLOTS

**Comment by yatli** [comment]
The problem seems to be pinned down to this pair:

```c++
    if (grid_col < devs[grid]->track_count) {
      SeqTrack *seq_track = mcl_seq.seq_tracks[i];
      mute_states[i] = seq_track->mute_state;
      seq_track->mute_state = SEQ_MUTE_ON;
    }
```

```c++
    if (grid_col < devs[grid]->track_count) {
      SeqTrack *seq_track = mcl_seq.seq_tracks[i];
      seq_track->mute_state = mute_states[i];
    }
```

@jmamma any ideas?

yatli:
`grid_col < devs[grid]->track_count` should guarantee we are within the bound though...

yatli:
awwww.

```
  void add_track(uint8_t track_number, uint8_t type, SeqTrack *tp) {
    seq_tracks[track_number] = tp;
    seq_tracks[track_number]->active = type;
  }
```

It ignores track type here

jmamma:
> awwww.
> 
> ```
>   void add_track(uint8_t track_number, uint8_t type, SeqTrack *tp) {
>     seq_tracks[track_number] = tp;
>     seq_tracks[track_number]->active = type;
>   }
> ```
> 
> It ignores track type here

what do you mean ignore? as in doesn't check ?

yatli:
MCLSeq::setup calls add_track(... EXT_TRACK_TYPE, ...) and overwrites the first MD track pointers, leaving the ext track pointers dangling.

jmamma:
ha. I see that now. 

i added him a while back, with no definitive end goal in mind.

yatli:
It's working now.

`Fuel: [|||--------------]`
AFK

jmamma:
same cooking 🍔

yatli:
The MCLClipboard bug may be related to around line 190. Should use pointers not EmptyTrack object directly.

jmamma:
> 190

example?

yatli:
`      header.update_model(x + col, empty_track.get_model(), empty_track.get_device_type());`

Will not use the virtual functions.

jmamma:
ack

yatli:
Almost there. (Do I sound like a Win10 installer)

yatli:
```
md kit okay
func_call: virtual bool MNMKit::fromSysex(MidiClass*)
mnm kit okay
```

yatli:
ok mnm models showing up in the grid.

the only problem is that kit number 0x7F wouldn't work at the moment.

@jmamma what's the workaround?

jmamma:
The old way. uint8_t kit = getCurrentKit, saveCurrentKit(kit), getBlockingKit(kit).

yatli:
Got a bad feeling that even kit dumps cannot go into MNM unless in sysex menu............

jmamma:
> Got a bad feeling that even kit dumps cannot go into MNM unless in sysex menu............

😱

yatli:
Even worse -- in the sysex menu it refuses to receive when it's playing!

yatli:
At first when I looked at the Ruin & Wesen legacy MNM implementation I thought, hmm, it's weird, how comes they have a very advanced implementation for MD but a half-baked solution for MNM?

Now I feel them. I really do.

jmamma:
The other alternative is CC transmission. 

If we have a generic driver for MIDI synths, this could be useful there too.

yatli:
Anyway, load/save MNM kit is complete (as well as this branch).
Save works even if it's playing, just chain write not working.

Next move: MMX

yatli:
Would you like to test this branch for regression (mostly about A4 then)?
For me I think it's ready for a merge.

jmamma:
You mean test A4Sound transmission ?

yatli:
Em, yeah that doesn't work :p

jmamma:
Calling it a night here.

Don't use up all your tank :rocket:

yatli:
Roger roger (starwar robot voice

jmamma:
I think we should merge this one in to ext_conv.

I have some changes to mcl_actions planned. and it would be easier using the updated code.

yatli:
I merged the latest changes in ext_conv. Project format changed?

jmamma:
Yes. I changed the grid layout. just create a new project. should fix

yatli:
alright. then it's good to merge now.

PR #129, opened 2020-08-16
restore vtable when load from grid

yatli:


jmamma:
I just did a quick glance over. Looking really good.

Will do a more thorough review later today.

avr/cores/megacommand/MCL/GridTrack.cpp
```diff
@@ -29,13 +59,13 @@ bool GridTrack::load_from_grid(uint8_t column, uint16_t row) {
...
 }
 
-bool GridTrack::store_in_grid(uint8_t column, uint16_t row) {
+bool GridTrack::store_in_grid(uint8_t column, uint16_t row, bool data) {
 
   DEBUG_PRINT_FN();
   bool ret;
 
-  uint32_t len = get_track_size();
-  ret = proj.write_grid(&(this->active), len, column, row);
```
jmamma, line 38: I don't think we should read/write the vtable pointer in to the project file.

Thoughts?

yatli: This is just for code brevity. Because I don't know how many vptrs are there, and how is the class organized, the only thing I can do (in a portable way) is to dump everything.

jmamma: so in other words. you're saying it would be unsafe to assume the 2byte vptr at the start of the object?

yatli: Exactly. The compiler might get creative here.

PR #128, opened 2020-08-10
MegaCom basic framework

yatli:
Work in progress. We can build a lot of things upon this:

```c
enum comserver_id_t {
  COMSERVER_EXTUI, // button input, ext screen etc.
  COMSERVER_FILESERVER, // simple FTP-ish protocol
  COMSERVER_EXTMIDI, // note in, CV out etc.
  COMSERVER_MAX,
  COMSERVER_DRAIN_INVALID = 0xFF
};
```

PR #130, opened 2020-08-22
Md sharedlock

yatli:


jmamma (commented):
woah. 

just did a first pass. everytime I think i've found a mistake, you've gone and done something clever.

will have another review a bit later.

jmamma:
Regarding the SeqStep page.

The last work I did on it was to enable editing one of [ step, slide, lock, mute ] masks

this was implemented using a  get_mask() function that would return pointer to relevant bitmask, based on the mask_type parameter selectable  from the (shift2) track menu.

```
 710 uint64_t *SeqPage::get_mask() {
 711   uint64_t *mask;
 712   if (opt_midi_device_capture == DEVICE_MD) {
 713     auto &active_track = mcl_seq.md_tracks[last_md_track];
 714     switch (mask_type) {
...
      }
       return mask;
  }
```

As we cannot refer to the bitmasks directly anymore, we can probably have a wrapper function for

set_trig(step, edit_type) 
clear_trig(step, edit_type)

where edit_type == mask_type.

and then some overloading for utiming, conditional.

yatli:
SeqStepPage:

      uint8_t pitch = active_track.get_track_lock(step, 0) - 1;

- what if the lock doesn't exist?
- here it says 0, but in MDSeqTrackData the param id is 1?

jmamma:
So 0 here is referencing the pitch parameter value (0), not lock_idx.

and it looks like

 active_track.get_track_lock(step, 0) returns 255 if there is no pitch lock on the step.

```
457 uint8_t MDSeqTrack::get_track_lock(uint8_t step, uint8_t track_param) {
458   uint8_t match = 255;
459   uint8_t c = 0;
460   // Let's try and find an existing param
461   for (c = 0; c < NUM_MD_LOCKS && match == 255; c++) {
462     if (locks_params[c] == (track_param + 1)) {
463       match = c;
464     }
465   }
466   if (match != 255) {
467     return locks[match][step];
468   }
469   return 255;
470 }
```

Following the logic, the tuning search will fail and seq_param4.cur will be set to 0

```
274       int8_t utiming = active_track.timing[step];
275       uint8_t pitch = active_track.get_track_lock(step, 0) - 1; <--- == 254
279       uint8_t note_num = 255;
280 
281       tuning_t const *tuning = MD.getModelTuning(MD.kit.models[last_md_track]);
282       if (tuning) {
283         for (uint8_t i = 0; i < tuning->len && note_num == 255; i++) {
284           uint8_t ccStored = pgm_read_byte(&tuning->tuning[i]);
285           if (ccStored >= pitch) { <--- this will never be satisfied 
286             note_num = i;
287           }
288         }
289         if (note_num == 255) {
290           seq_param4.cur = 0;  <---- note_num default value.
291         } else {
292           seq_param4.cur = note_num;
293         }
294         seq_param4.old = seq_param4.cur;
295       }
```

yatli:
By "idx 1" I was refering to this:

void MDSeqTrack::set_track_pitch(uint8_t step, uint8_t pitch) {
  set_track_locks(step, 1, pitch);
}

jmamma:
is there a bug/inconsistency there? 

lock ids should be shifted by 1. 

as 0 means no lock.

yatli:
set track pitch uses param id 1 for set, but the SeqStepPage uses param id 0 for get. 🤔

yatli:
```
  FORCED_INLINE() uint8_t get_lockidx(uint8_t step, uint8_t lock_idx) const {
    uint8_t mask = 1 << lock_idx;
    uint8_t rmask = ~(mask - 1);
    if (steps[step].is_lock(lock_idx)) {
      auto idx = get_lockidx(step) + popcount(steps[step].locks & rmask);
      return idx;
    } else {
      return 255;
    }
  }
```

A problem with this routine is that 255 is a valid index... 🤦‍♀️

Edit: alright, let's go uint16_t and return NUM_MD_LOCK_STEPS to indicate not found.

yatli:
```
void MDTrack::scale_seq_vol(float scale) {
  for (uint8_t c = 0; c < NUM_MD_LOCKS; c++) {
    if ((seq_data.locks_params[c] == MODEL_LFOD + 1) ||
        (seq_data.locks_params[c] == MODEL_VOL + 1)) {
      for (uint8_t n = 0; n < NUM_MD_STEPS; n++) {
        auto idx = seq_data.get_lockidx(n, c);
        if (idx != NUM_MD_LOCK_SLOTS) {
          seq_data.locks[idx] = (uint8_t)(scale * (float)(seq_data.locks[idx] - 1)) + 1;
          if (seq_data.locks[idx] > 127) {
            seq_data.locks[idx] = 127;
          }

        }
      }
    }
  }
}
```

          if (seq_data.locks[idx] > 127) {
            seq_data.locks[idx] = 127;
          }

should be 128 because it's +1?

jmamma:
the shifting lock_params and lock_values by +1 is not a great solution

ideally the NULL slot test should be if val == 255 not if val == 0.

This was maintained this way for legacy reasons.

yatli:
If all lock values are 7-bit, we can just use the MSB for testing.

if (val & 0x80) ...

yatli:
// !! Note track_param is param_id, not lock index
  bool set_track_locks(uint8_t step, uint8_t track_param, uint8_t velocity);
  // !! Note track_param is lock index, not param_id
  uint8_t get_track_lock(uint8_t step, uint8_t track_param);

Okay I figured it out...

yatli:
It builds now, but nothing works yet.

jmamma:
a familiar story 😆

yatli:
I need to stop now. So much procrastination on the day job...

But it seems I can't. (panic)

jmamma:
all good! i'll circle round to this one soon.

yatli:
most probably this implementation is going to be fine. lock insertion/deletion tested. rlck tested (without the optimization... thought it'll be needed)

yatli:
~~could you take a look at SeqParamPage? handleEvent seems a little broken.~~

edit: done

jmamma:
building now...

jmamma:
```
/Applications/Arduino.app/Contents/Java/hardware/valence/avr/cores/megacommand/MCL/MDSeqTrack.h: In member function 'void MDSeqTrack::get_mask(uint64_t*) const':
/Applications/Arduino.app/Contents/Java/hardware/valence/avr/cores/megacommand/MCL/MDSeqTrack.h:57:12: error: expected '(' before 'constexpr'
         if constexpr (mask_type == MASK_LOCK) {
```

yatli:
uh oh...

yatli:
you'll need a newer toolchain that supports c++17 for this

yatli:
.. or let me refactor this a little bit

jmamma:
hmm let me see what the latest arduino binaries are packaged with.

jmamma:
no wonder we never get the same binaries. on build.

jmamma:
from arduino changelog.
```
[core]
* Updated AVR core to 1.8.3:
  - Wire class now supports timeout
  - Upgraded avr-gcc to 7.3.0-atmel3.6.1-arduino7
```

yatli:
alright you're good to go now.
param-c and param-d are online.

jmamma:
wow great stuff. got 8 param locks working.

tune on seq_step_page is setting decay not ptc. so assume +1 issue.

jmamma:
nice!

i think i can take it from here.

When I have time,

I'd like to refactor this ugly beast.
```
 560 
 561 void SeqPage::draw_mask(const uint8_t offset, const uint64_t &pattern_mask,
 562                         const uint8_t step_count, const uint8_t length,
 563                         const uint64_t &mute_mask, const uint64_t &slide_mask,
 564                         const bool show_current_step) {

**Comment by yatli** [comment]
Note, `const uint64_t &` will be a lot faster than passing the parameter by value.

**Comment by yatli** [comment]
there might be a bug in the lock slot manipulation code -- editing one lock sometimes result in changing the next one (so the locks are not on-beat)

**Comment by jmamma** [comment]
> there might be a bug in the lock slot manipulation code -- editing one lock sometimes result in changing the next one (so the locks are not on-beat)

yeah, i noticed some issue setting parameter locks from step page. will investigate.

**Comment by jmamma** [comment]
```
218           switch(mask_type) {
219             case MASK_LOCK:
220               // TODO what does it mean to "set lock mask"?
221               break;
```

You could previously disable locks for a step, without deleting the parameter values.

Not sure if new system works this way?

**Comment by yatli** [comment]
You could previously disable locks for a step, without deleting the parameter values.

Uh oh...

When we remove the locks, they get overwritten by memmove.

**Comment by jmamma** [comment]
When we remove the locks, they get overwritten by memmove.

Can this be decoupled ?

cond_plock = false, but don't flush bitfield.

**Comment by yatli** [comment]
Two options:

- We have one spare bit in the descriptor. Can assign to `disable_locks`
- When storing the lock values, don't +1, so that we can use the MSB as "lock disable".

How can we use this feature? In SeqParamPage, or in SeqStepPage, with "edit locks"?

**Comment by jmamma** [comment]
```
When we remove the locks, they get overwritten by memmove.

Can this be decoupled ?

cond_plock = false, but don't flush bitfield.
```

howcome decoupling is not possible?

**Comment by yatli** [comment]
`cond_plock = false` means conditional doesn't govern plocks, not "condition of plock is false" :D :D

**Comment by jmamma** [comment]
oh right, I did work that out earlier. half asleep today.

**Comment by yatli** [comment]
> condition of plock is false

The MSB solution will satisfy exactly this requirement

**Comment by jmamma** [comment]
Let's use up the extra bit in the descriptor

ie

```
 41   uint8_t
 42       locks; // <-- bitfield of 8 locks in the current step, first lock is lsb
         bool  lock : 1;
 43   bool trig : 1;
 44   bool slide : 1;
 46   uint8_t cond_id : 4;
```

**Comment by yatli** [comment]
But the question is, when do we completely erase a lock? When memory is full and we do a "garbage collection"-ish sweep?
A mini file system for locks it becomes :)

**Comment by jmamma** [comment]
user needs to delete the locks from the step from shift menu.  or clear the pattern.

(this is current behaviour) 

--

using the MSB in lock value means you have potentially 8 x MSB to generate 1x lock bit mask.

**Comment by jmamma** [comment]
> user needs to delete the locks from the step from shift menu. or clear the pattern.
> 
> --
> 
> using the MSB in lock value means you have potentially 8 x MSB to generate 1x lock bit mask.

**Comment by yatli** [comment]
What will happen if:

- Lock memory is full
- Some locks are disabled
- The user wants to input a new lock

**Comment by jmamma** [comment]
oled_display.textbox("LOCKS","FULL");

**Comment by yatli** [comment]
But it's not full meow (

**Comment by yatli** [comment]
Another question.

Say there are 3 locks on a step, and now we disable the locks.
When later the user inputs a lock value, does that mean all 3 locks are activated again?

If yes: go with `bool locks_enabled:1;`
If no: go with `MSB`

**Comment by jmamma** [comment]
Yes all locks become enabled again.

with MSB you would need a lock mask per parameter to enable/disable individually, which we're not going to do.

**Comment by yatli** [comment]
~~Done.~~

Edit: not yet :p

**Comment by yatli** [comment]
Done this time.

**Comment by yatli** [comment]
@jmamma it's behaving like a sane sequencer. ready for some testing now.

**Comment by jmamma** [comment]
cool. im heading out for a bit of exercise. will test in a few hours.

one other thing that needs doing is that step input is hardcoded for trig_pattern , and should work to control the various masks types.

**Comment by yatli** [comment]
> a bit of exercise

I'll need some too when I recover from the cold (just cold.)

**Comment by yatli** [comment]
> one other thing that needs doing is that step input is hardcoded for trig_pattern , and should work to control the various masks types.

Done.

**Comment by jmamma** [comment]
you're too quick. it thought you were taking a break :p

**Comment by jmamma** [comment]
slides are broken-ish.

**Comment by jmamma** [comment]
```
204   for (uint8_t i = 0; i < NUM_MD_LOCKS; i++) {
205     if (locks_params[i] && (steps[i].locks & cur_mask)) {
206       find_mask |= cur_mask;
207     }
208     cur_mask <<= 1;
209   }
```

Why do you need to calculate a find mask for the current step.
Wouldn't the search mask equal to steps[current_step].locks  ?

also steps[i].locks is an error?

**Comment by yatli** [comment]
> Wouldn't the search mask equal to steps[current_step].locks ?

because locks_params[i] must be greater than 0

> also steps[i].locks is an error?

~~nope :p~~
edit: ah I see. I saw it and it made me feel uncomfortable for a moment, and then I forgot D:

sliding works for wrap-around case but not for the easy one... 🤔

**Comment by yatli** [comment]
fixed it.

**Comment by jmamma** [comment]
```
Wouldn't the search mask equal to steps[current_step].locks ?
because locks_params[i] must be greater than 0
```

Okay, so this takes care of the scenario:

(steps[current_step].locks >> a) & 1 == true
&&
locks_params[a] == 0 

Where user might disable lock param, but the locks are still there.

**Comment by yatli** [comment]
Yeah. By turning off the target knob, MCL can save some cycles searching for the locks.

**Comment by jmamma** [comment]
awesome.

just had a test. looking good so far.

**Comment by jmamma** [comment]
We don't really need the fastest algorithm for get_*_mask();

We can save 100 bytes per function if we do something like:
 
```
void MDSeqTrack::get_slide_mask(uint64_t *_pmask) const
{
    *_pmask = 0;
    for(int i=0;i<NUM_MD_STEPS; i++) {
      if ((steps[i].slide)) {
      SET_BIT64_P(_pmask, i); 
      }   
    }   
}
```

**Comment by yatli** [comment]
~~premature~~ pre-birth optimization 😅

thought it'd be helpful for boosting the GUI smoothness a bit.
after second thought, just drawing a single character on screen would erase the benefit

**Comment by jmamma** [comment]
> ~premature~ pre-birth optimization 😅
> 
> thought it'd be helpful for boosting the GUI smoothness a bit.
> after second thought, just drawing a single character on screen would erase the benefit

a lot of the firmware bloat is coming from small auxiliary functions. so any little savings we can make in bytes is helpful, when performance is not critical.

**Comment by yatli** [comment]
understood.

**Comment by jmamma** [comment]
mute/oneshot masks not displaying anymore. was this working earlier ?

**Review by yatli — COMMENTED** [review]
LGTM

**Comment by yatli** [comment]
> mute/oneshot masks not displaying anymore. was this working earlier ?

Can confirm. Not working on this branch. Perhaps I set the wrong mask.

File: avr/cores/megacommand/MCL/MCLClipBoard.cpp [hunk 1]
```diff
@@ -66,11 +66,9 @@ bool MCLClipBoard::copy_sequencer_track(uint8_t track) {
   }
 
   if (track < NUM_MD_TRACKS) {
-    memcpy(&(md_track->seq_data), &mcl_seq.md_tracks[track],
+    memcpy(md_track->seq_data.data(), mcl_seq.md_tracks[track].data(),
```
jmamma, line 67: hmm. what's this?

avr/cores/megacommand/MCL/MDSeqTrackData.h
```diff
@@ -22,31 +28,69 @@ class MDSeqTrackData_270 {
...
-    memset(&locks_params, 0, NUM_MD_LOCKS);
-    pattern_mask = 0;
-    lock_mask = 0;
-    memset(&conditional, 0, NUM_MD_STEPS);
-    memset(&timing, 0, NUM_MD_STEPS);
+  MDSeqStepDescriptor steps[NUM_MD_STEPS];
+
+  // get the pointer to the data chunk.
+  // useful to skip the vtable
+  void* data() const { return locks; }
```
jmamma: and this ? should the return be a pointer ?

avr/cores/megacommand/MCL/MCLClipBoard.cpp
jmamma: copy/paste track is currently broken fyi.

yatli: prevent unaligned copy from something with vtable to something without one.
MDTrack gets MDSeqTrackData (no vtable)
MDSeqTrack has one, coming from SeqTrack 😱

avr/cores/megacommand/MCL/MDSeqTrackData.h
yatli: ah, yes... wondering how it gets through the type check.

avr/cores/megacommand/MCL/MCLClipBoard.cpp
jmamma: ah okay. so we want to skip copying the vtable. got you.

jmamma:
```
128     memcpy(mcl_seq.md_tracks[track].data(), md_track->seq_data.data(),
129            sizeof(md_track->seq_data));
```

isn't sizeof going to  be incorrect then?

yatli: md_track->seq_data has the correct size (😱)

avr/cores/megacommand/MCL/MDSeqTrackData.h
yatli: here locks _IS_ a pointer.

jmamma:
``` 
 51 class MDSeqTrackData {
 52 public:
 53   uint8_t locks[NUM_MD_LOCK_SLOTS];
```
?

jmamma: ah because it's an array.

jmamma:
Forgive my confusion but:

788545d

Are we now no longer using + 1 for both lock_params and locks

--

If so, there's a few areas of code that still need cleanup.

avr/cores/megacommand/MCL/MCLClipBoard.cpp
jmamma: not sure why this has broken copy/paste.

yatli:
ah, yes. I've cleaned up a few places but there could be more.

avr/cores/megacommand/MCL/RAMPage.cpp
```diff
@@ -198,61 +199,64 @@ bool RAMPage::slice(uint8_t track, uint8_t linked_track) {
...
   bool send_params = false;
-  mcl_seq.md_tracks[track].clear_track(clear_locks, send_params);
 
-  mcl_seq.md_tracks[track].locks_params[0] = ROM_STRT + 1;
-  mcl_seq.md_tracks[track].locks_params[1] = ROM_END + 1;
+  auto &trk = mcl_seq.md_tracks[track];
+  auto &ln_trk = mcl_seq.md_tracks[linked_track];
+  trk.clear_track(clear_locks, send_params);
+
+  trk.locks_params[0] = ROM_STRT + 1;
```
jmamma, line 206: fix

avr/cores/megacommand/MCL/RAMPage.cpp
```diff
@@ -198,61 +199,64 @@ bool RAMPage::slice(uint8_t track, uint8_t linked_track) {
...
+  trk.locks_params[1] = ROM_END + 1;
```
jmamma, line 207: fix

avr/cores/megacommand/MCL/MDTrack.cpp
```diff
@@ -104,19 +103,19 @@ void MDTrack::place_track_in_sysex(uint8_t tracknumber) {
...
-            seq_data.locks[c][n] =
-                (uint8_t)(scale * (float)(seq_data.locks[c][n] - 1)) + 1;
-            if (seq_data.locks[c][n] > 127) {
-              seq_data.locks[c][n] = 127;
-            }
+  for (uint8_t n = 0; n < NUM_MD_STEPS; n++) {
+    auto idx = seq_data.get_lockidx(n);
+    for (uint8_t c = 0; c < NUM_MD_LOCKS; c++) {
+      if (seq_data.steps[n].is_lock(c)) {
+        if ((seq_data.locks_params[c] == MODEL_LFOD + 1) ||
```
jmamma, line 109: fix

avr/cores/megacommand/MCL/MDTrack.cpp
```diff
@@ -104,19 +103,19 @@ void MDTrack::place_track_in_sysex(uint8_t tracknumber) {
...
+            (seq_data.locks_params[c] == MODEL_VOL + 1)) {
```
jmamma, line 110: fix

avr/cores/megacommand/MCL/MDTrack.cpp
```diff
@@ -104,19 +103,19 @@ void MDTrack::place_track_in_sysex(uint8_t tracknumber) {
...
+          seq_data.locks[idx] =
```
jmamma, line 111: fix

jmamma (commented):
marking these locations for offset fix.

avr/cores/megacommand/MCL/MCLClipBoard.cpp
jmamma:
```
func_call: bool MCLClipBoard::copy_sequencer_track(uint8_t)
func_call: bool MCLClipBoard::open()
failed write
6
6
func_call: bool MCLClipBoard::paste_sequencer_track(uint8_t, uint8_t)
func_call: bool MCLClipBoard::open()
failed read
```

unrelated to this change.

yatli:
sorry I was in a meeting and haven't made it clear:

the locks +1 offset are removed, but lock_params are not

yatli:
pls revert 5b2c829

jmamma:
> pls revert [5b2c829](https://github.com/jmamma/MIDICtrl20_MegaCommand/commit/5b2c8294713c9c8b90560cea74c53bc412437228)

?

jmamma:
> the locks +1 offset are removed, but lock_params are not

ack.

avr/cores/megacommand/MCL/MDSeqTrack.cpp
```diff
@@ -700,75 +785,119 @@ void MDSeqTrack::merge_from_md(uint8_t track_number, MDPattern *pattern) {
...
+    if (step->locks[a]) {
+      ++delta;
+      new_locks |= 1;
+    }
+    new_locks <<= 1;
   }
-  if (step->slide_mask) {
-    SET_BIT64(slide_mask, n);
+
+  if (delta < 0) { // shrink
```
jmamma: You've added the code for shifting the array, but I think you've missed the actual lock copy/insert?

jmamma: actually, is the array shifting needed at all.

can we just recycle

 bool MDSeqTrack::set_track_locks_i(uint8_t step, uint8_t lockidx,
545                                    uint8_t value) {

to save a few bytes?

yatli: you're right :D just lay down the lock data at idx+nlock, the space is properly reserved.

jmamma: will do

avr/cores/megacommand/MCL/MDSeqTrackData.h
```diff
@@ -48,6 +37,14 @@ class MDSeqStepDescriptor {
   bool is_lock(const uint8_t idx) const { return is_lock_bit(idx) && locks_enabled; }
 };
 
+class MDSeqStep {
```
yatli, line 40: nice.

avr/cores/megacommand/MCL/MDSeqTrack.cpp
yatli: > can we just recycle....

set_track_locks only set one lock at a time, so it means resizing the array up to 8 times. so the question is, do we want to pay the price.

.... or, make set_track_locks_i work for multiple locks at once?

jmamma: I suppose if we implement copying multiple steps (copy page for example) it could have more of an impact.

also, what are the other scenario where we will be setting multiple locks together?
In live record we would need to have some buffer mechanism in CC callback to make use of this.

Furthermore, the majority of the work is the 
```
556     memmove(locks + lock_slot + 1, locks + lock_slot,
557             NUM_MD_LOCK_SLOTS - lock_slot - 1);
```

how many instructions is this? for 256 byte array it might not be that bad afterall.

jmamma: Thinking a little larger...

Eventually, we will need to generalise these routines so that they can be shared with the ExtSeqTrack data structures

Recording notes in realtime would probably benefit from having a multiple insert routine.

yatli: How do we eval the performance of RLCK? I tried a little bit and was surprised it worked well enough for me as-is, without the circular insertion buffer we discussed.

jmamma: The important performance metric is latency. 

If there is a long wait to insert the next lock, it may get applied at the wrong sequencer position.

It might be sufficient to show that insertion time is much shorter than sequencer resolution.  i.e 
t_insertion < MidiClock.div192_time

jmamma (commented):
🙄

yatli:
just sayin'

yatli:
ready to pop the stack?

jmamma:
send_tracks_to_devices is not working in ext_conv branch.

was aiming to troubleshoot that first, before pushing more on to the commit history.

yatli:
We need a better diagnostic page (overlay).

jmamma:
What metrics do you have in mind?

yatli:
Not yet, I'm thinking about a mechanism, just like how we hack the textbox into the display routine.
Instead of DEBUG_DUMP, we can implement DIAG_DUMP(val, slot) and DIAG_MSG(str).

Once these macros are called, the diag page becomes active, and will be displayed as an overlay when oled_display.display() is called.

The panel is split in two halves.
Left = 6 dump slots
Right = 6 lines of diag messages

Thoughts?

jmamma:
Sure.. we can have a separate macro for DIAG_MODE

You should be able to follow the textBox code to inject another overlay on top.

yatli:
Yep. :)

Also, recommend cherry-picking the debug formatter from https://github.com/jmamma/MIDICtrl20_MegaCommand/pull/128/commits/a1ce9bfaec395a62256e3f00548fc131a2e2133e

jmamma:
we can move over to megacom in a little while.

yatli:
okay, let me kick off this one.
been wanting this for a long time

yatli:
Some rlck benchmark result.

Step=64, all trigs on.
Activate rlck and hit play, turn the knobs.

MCLSeq::seq() time: around 750us, spikes to 2000us
set_track_locks time: depends on the lock index, and the number of occupied slot count:

- 1 lock param active: ~20us
- 6 lock param active, later locks: ~160us (!)
- 8 lock param active, half way in the pattern: ~80us

Not bad, but could be improved indeed.

PR #134, opened 2020-09-21
Ext new format

yatli:
@jmamma we still need to adapt the editor to use it.

~~but given the current situation we may need to first casecade back to piano roll branch.~~

edit: since there are no new stuff in piano_roll we can just proceed on this branch..


yatli:
The ext track size is reduced from 2000+ bytes to 1600+ bytes, also gaining the following capabilities:

- 8 lock targets
- up to 16 events per step (lock + note)
- individual utiming for each event
- lock engage/disengage behavior: a lock-engage event is sticky, like conventional automation in a piano roll. Use another lock-disengage event to recall the original lock value (the plock behavior)

The actual number of slots to keep events is still 512 -- we're still able to fill the track with 4x128 midi notes. But now locks and notes share the same event storage.

Wrt the lock engage/disengage behavior: I designed it so that we overcome the inconsistency of long notes vs. single-step p-lock.
The actual behavior is now left to the track editor -- both sticky/plock behavior can be described with this storage model.

yatli:
Afk for now. @jmamma can you take a look at the implementation and see if you can work out SeqExtStepPage?

yatli:
I'll unpack my A4 when I get home...

jmamma:
https://en.wikipedia.org/wiki/Nibbles_(video_game)

First video game I played on the PC.

avr/cores/megacommand/CommonTools/NibbleArray.h
```diff
@@ -0,0 +1,62 @@
...
+	void shift_right(uint8_t len) {
+		if (len == 0) return;
+		NibbleArray<N> that = *this;
+		for(uint8_t i = 1; i < len; ++i) {
+			set(i, that.get(i-1));
+		}
+		set(0, that.get(len-1));
+	}
+
+	void reverse(uint8_t len) {
```
jmamma, line 54: This implementation won't work. Once you get to i/2 you have already overwritten the remaining values.

avr/cores/megacommand/CommonTools/NibbleArray.h
```diff
@@ -0,0 +1,62 @@
...
+		for(uint8_t i = 0; i < len; ++i) {
+			set(i, that.get(i+1));
+		}
+		set(len, that.get(0));
+	}
+
+	void shift_right(uint8_t len) {
+		if (len == 0) return;
+		NibbleArray<N> that = *this;
+		for(uint8_t i = 1; i < len; ++i) {
```
jmamma, line 48: this loop needs be reversed?

avr/cores/megacommand/MCL/ExtSeqTrack.cpp
```diff
@@ -31,6 +29,63 @@ void ExtSeqTrack::re_sync() {
...
+uint16_t ExtSeqTrack::add_event(uint8_t step) {
+  uint8_t u = timing_buckets.get(step); 
+  if (15 == u || event_count >= NUM_EXT_EVENTS) {
+    // bucket full or track full
+    return 0xFFFF;
+  }
+  ++event_count;
+  timing_buckets.set(step, u+1);
+  uint16_t idx, end;
+  locate(step, idx, end);
```
jmamma, line 56: ~idx, end should be a &idx, &end~

avr/cores/megacommand/CommonTools/NibbleArray.h
yatli: "that" is copied

avr/cores/megacommand/MCL/ExtSeqTrack.cpp
yatli: yeah let's go modern

avr/cores/megacommand/CommonTools/NibbleArray.h
yatli: Nope, also copied.

avr/cores/megacommand/CommonTools/NibbleArray.h
jmamma: (jmamma) c+ -> (yatli) c++

jmamma: (jmamma) c+ -> (yatli) c++

jmamma:
Quick pass 
70d75c6

Just tried recording the A4 from chromatic page and doesn't sound like anything is being recorded.

Need to refuel. 🍔

jmamma:
Crash:

Chromatic page.. record a few notes then press shift1 (page_select)

jmamma:
```
Sending note
pitch = 21
note_idx = 65535
note_num = 45
step_count = 1
"here" = here
note_idx = 37896
```

Bad idx ?


```
340 
341   if (note_idx == 0xFFFF) {
342     DEBUG_DUMP(note_idx);
343     DEBUG_DUMP(note_num);
344     DEBUG_DUMP(step_count);
345  
346           note_idx = add_event(step_count);
347   } 
348   DEBUG_DUMP("here");
349   DEBUG_DUMP(note_idx);
350   if (note_idx != 0xFFFF) {
351    events[note_idx].is_lock = false;
352     events[note_idx].event_on = true;
353     events[note_idx].event_value = note_num;
354     events[note_idx].cond_id = condition;
355     events[note_idx].micro_timing = utiming;
356   }
357 }
```

jmamma:
looks like missing return.

jmamma:
Got some note ouput now. but something still not right.

jmamma:
Handing back over to you 💤

avr/cores/megacommand/MCL/ExtSeqTrack.cpp
```diff
@@ -31,6 +29,69 @@ void ExtSeqTrack::re_sync() {
...
+  uint8_t u = timing_buckets.get(step);
+  if (15 == u || event_count >= NUM_EXT_EVENTS) {
+    // bucket full or track full
+    return 0xFFFF;
+  }
+  ++event_count;
+  timing_buckets.set(step, u + 1);
+  uint16_t idx, end;
+  locate(step, idx, end);
+  memmove(events + end + 1, events + end,
```
jmamma, line 57: Is this shifting correct?

I thought the new event is being placed at events + idx?

yatli: should be good. I'm inserting at the end of the current bucket.

yatli: The memmove size can be reduced to sizeof(ext_event_t) * (event_count - end - 1) though

avr/cores/megacommand/MCL/ExtSeqTrack.cpp
```diff
@@ -44,27 +105,38 @@ void ExtSeqTrack::seq() {
...
+    uint16_t ev_idx, ev_end;
+
+    // Locate CURRENT
+    locate(step_count, ev_idx, ev_end);
+
+    // Go over CURRENT
+    for (; ev_idx != ev_end; ++ev_idx) {
+      auto u = events[ev_idx].micro_timing;
+      if (u >= timing_mid && u - timing_mid == mod12_counter) {
+        handle_event(ev_idx);
```
jmamma, line 123: fyi: when lock events are added, they need to be sent before note_on/off .
This ensures machine params are updated before sound trigger.

yatli: ah, yes.. need to do it in the editor part

yatli: also, velocity locks should be treated specially because that's part of the midi note event.

jmamma:
Just doing some more testing.

If I record an arbitrarily long sequence greater than 1 note, the first note is never played back.

jmamma:
A few things. timing_buckets is not initialized on creation? 

```
121   void locate(uint8_t step, uint16_t& ev_idx, uint16_t& ev_end) {
122     ev_idx = 0;
123     ev_end = timing_buckets.get(step_count);
124     for (uint8_t i = 0; i < step_count; ++i) {
125       ev_idx += timing_buckets.get(i);
126     }
127 
128     ev_end += ev_idx;
130   }
```

Edit: ev_end is max range, this value is never used as index.

--

On SeqExtStepPage the note detection routine only seems to run when the ISR triggers note out messages on port 2. This is extremely weird. The note detection is not working correctly, but it flashing bogus notes when note_on/off is transmitted.

avr/cores/megacommand/MCL/ExtSeqTrack.cpp
```diff
@@ -31,6 +29,69 @@ void ExtSeqTrack::re_sync() {
...
+  // step is at least 1, move back to get current step
+  --step;
+  timing_buckets.set(step, bucket - 1);
+  memmove(events + index, events + index + 1,
+          sizeof(ext_event_t) * (NUM_EXT_EVENTS - index - 1));
+  --event_count;
+}
+
+uint16_t ExtSeqTrack::add_event(uint8_t step) {
+  uint8_t u = timing_buckets.get(step);
```
jmamma, line 49: shouldn't this line, be placed after the memmove.

The locate(step,idx,end) command will return u + 1

avr/cores/megacommand/MCL/ExtSeqTrack.cpp
```diff
@@ -31,6 +29,69 @@ void ExtSeqTrack::re_sync() {
...
+  if (15 == u || event_count >= NUM_EXT_EVENTS) {
+    // bucket full or track full
+    return 0xFFFF;
+  }
+  ++event_count;
+  timing_buckets.set(step, u + 1);
```
jmamma: shouldn't this line, be placed after the memmove.

The locate(step,idx,end) command will set end = u + 1

avr/cores/megacommand/MCL/ExtSeqTrack.h
```diff
@@ -120,11 +120,13 @@ class ExtSeqTrack : public ExtSeqTrackData, public SeqTrack {
 
   void locate(uint8_t step, uint16_t& ev_idx, uint16_t& ev_end) {
     ev_idx = 0;
-    ev_end = timing_buckets.get(step_count);
-    for (uint8_t i = 0; i < step_count; ++i) {
+    ev_end = timing_buckets.get(step);
+    for (uint8_t i = 0; i < step; ++i) {
```
jmamma, line 123: 😱 @yatli

jmamma:
Ext live record is now working 🚀 🌔 

ExtSeqStepPage is showing the recorded notes 👍

PR #138, opened 2020-11-23
Sdfat v2

jmamma:
V2 simplifies directory traversal? 

vwd() is now private, and instead of vwd()->rewind. SD.chdir() should do.

However, we still use SD.vwd()->getName(temp_entry, 16);  Doesn't appear to be an equivalent mechanism?
 
https://github.com/greiman/SdFat/blob/master/examples/DirectoryFunctions/DirectoryFunctions.ino

Workaround was to make vwd() public again.

--

V2 compiles but there is a link error.

```
/var/folders/0q/mn27032j7bqch7d_krf2603c0000gn/T//ccBppv7c.ltrans2.ltrans.o: In function `FatFile::cacheDirEntry(unsigned char)':
<artificial>:(.text+0x5a9c): undefined reference to `FsDateTime::callback'
<artificial>:(.text+0x5aa0): undefined reference to `FsDateTime::callback'
/var/folders/0q/mn27032j7bqch7d_krf2603c0000gn/T//ccBppv7c.ltrans2.ltrans.o: In function `FatFile::write(void const*, unsigned int)':
<artificial>:(.text+0x6cd0): undefined reference to `FsDateTime::callback'
<artificial>:(.text+0x6cd4): undefined reference to `FsDateTime::callback'
/var/folders/0q/mn27032j7bqch7d_krf2603c0000gn/T//ccBppv7c.ltrans2.ltrans.o: In function `FatFile::open(FatFile*, fname_t*, unsigned char)':
<artificial>:(.text+0x7f52): undefined reference to `FsDateTime::callback'
/var/folders/0q/mn27032j7bqch7d_krf2603c0000gn/T//ccBppv7c.ltrans2.ltrans.o:<artificial>:(.text+0x7f56): more undefined references to `FsDateTime::callback' follow
collect2: error: ld returned 1 exit status
make: *** [main.elf] Error 1
```

FsDateTime::callback is a user-defined TimeDate callback. For using GPS timestamps and the like.


jmamma:
Got it to compile by commenting out references to FsDateTime::callback
MCL running okay.

Compile size:

V1:  224334
V2: 224370 (using standard config options, no exfat support)

wonder if performance is better

jmamma:
~project creation appears to be quite a bit faster on v2.~

edit:

stopwatch says they are more or less the same.

jmamma:
row write extremely slow. hmm

edit: bug with project load/creation?

jmamma:
Getting SDFat 2.0 to link correctly simply requires adding "SdFat/common" to the DIRS parameter in Makefile.

PR #148, opened 2020-12-12
Assign machine

jmamma:


jmamma:
Saves around 700bytes with this implementation.

Something wrong with parameter transmission/recv

jmamma:
Unstable on dev branch. Because SYSEX requests lock up the ISR, sequencer tracks will drift apart.
Ironically this is more stable on defer_rendering branch which uses a side channel buffer for all sequencer data and no blocking during sysex.

jmamma:
Shit.

We have to PATTERN_INDEX variables.

On a hunch, i changed the kit of some of my patterns that have KIT02 loaded. now undo is switching to their kit.

PR #156, opened 2021-03-23
implement locks via encoder interface

jmamma:


avr/cores/megacommand/MCL/SeqStepPage.cpp
```diff
@@ -324,14 +337,52 @@ bool SeqStepPage::handleEvent(gui_event_t *event) {
...
+        for (uint8_t n = 0; n < NUM_MD_TRACKS; n++) {
+          if (note_interface.notes[n] == 1) {
+            active_track.set_track_locks(n, param, MD.kit.params[last_md_track][param]);
+            trig_interface.ignoreNextEvent(key);
+          }
+        }
+      }
+      send_locks(step);
+      break;
+    case MDX_KEY_YES:
```
jmamma, line 379: @yatli can you look at this switch statement. MDX_KEY_YES doesn't execute.

feels like i've made an obvious mistake ??

jmamma: patch below fixes. do I need to re-read the C/C++ manual ?

PR #159, opened 2021-04-06
MegaCom and Sample management

yatli:
Continuation of:

https://github.com/jmamma/MIDICtrl20_MegaCommand/pull/104
https://github.com/jmamma/MIDICtrl20_MegaCommand/pull/128

Those branches are too outdated but I have this `comm_wav_merge` that I kept synchronizing to master.

@jmamma With the latest `enhanced_gui` code merged, the PROGMEM limit is reached so I have to drop the wave editor -- but this is not to abandon this functionality -- we can move this to MD.


yatli:
- .spk sample packs in SD-Drive
  - [x] Push
- .syx in SoundBrowserPage
  - [ ] Send (NOT WORKING)
- SampleManager with gateway 0x34
  - [x] List, implementation in FileBrowserPage
  - [ ] Rename, erase, copy/paste
- Better .wav file support
  - [x] extended attribute
  - [x] multi-channel wav files
  - [ ] smpl chunk
    - [x] loop points
    - ~~[ ] root node~~
    - [ ] slices
- ~~WavEdit~~

jmamma:
I'm okay with stripping out the wav editor for now. 

Unless you think otherwise, MegaCom should not be added to next MCL release, given the scope of changes to GUI and potential for bugs. 

We can do a small MCL release shortly after

yatli:
> Unless you think otherwise, MegaCom should not be added to next MCL release, given the scope of changes to GUI and potential for bugs.

So is the `wav` branch, so I think they can be coordinated.
I cannot maintain two old branches at once T.T
So my plan is to hang this PR out and absorb anything that comes into master along the way.

PR #162, opened 2021-04-10
Comp

yatli:


yatli:
Note to tomorrow me: need something like `[FieldOffset(...)]` for C.
Possible to achieve with unions and anonymous pads.
Need the script to align symbol names with sizes and types.

PR #191, opened 2024-07-16
MIDI Machine sequencing on port2 via MDSeqTrack, preliminary

jmamma:
Compiles, completely untested.

Not sure if note length is correct.

Arpeggiator code needs to be looked at as well for the MIDI Machine type.


jmamma:
So this is partly working. It'll trigger notes, chords and velocity is working including parameter locks.

However, note length is not doing anything. and not sure if the CCs are working.

PR #197, opened 2026-07-21
Add MIDI CC manual step-advance mode for Machinedrum tracks

viktorakos95:
I'm not a programmer. I built this with help from AI (Claude), based on my own idea and tested on my own hardware. Wanted to be upfront about that before anyone reviews it.

The idea: I wanted a way to trigger the Machinedrum sequencer step by step using a MIDI CC message, instead of always following the clock. So when this is turned on, sending a CC advances the sequencer one step. The clock itself doesn't stop, everything else, other gear, LFOs, stays synced normally. It's just the Machinedrum step advance that gets taken over by the CC.

I use this with an envelope follower, so a drummer or instrument playing into it can push the sequencer forward step by step. But the options are many. You could advance with the push of a button, a footswitch, or even an encoder to create experimental sounds. Really anything that can send a MIDI CC.

New options show up in the SEQ menu, right under QUANT: turn it on/off, choose the CC number, choose which MIDI port it listens on. It only listens on MIDI2 or USB, never MIDI1, since MIDI1 is where the Machinedrum sends its own CC messages and I didn't want the two getting mixed up.

It saves per project, so different projects can have this off, or on with different CC numbers/ports. I also made sure older project files still open fine without this feature confusing them.

One extra thing: while testing I ran into a memory limit on my MegaCommand DIY board, so I turned off the "WAV Designer" feature to free up space for this. It's not something I use, so I disabled it rather than trim down the new feature. There might be other ways to free up the memory for this to fit that I'm not aware of, so open to suggestions there.

I tested this on my own MegaCommand DIY hardware. Menu works, CC triggers steps correctly, clock keeps running for other gear, and settings save/load correctly per project across reboots and restarts.

Open to feedback, happy to adjust anything.


jmamma:
Hi @viktorakos95 

The last year or so I have been refactoring the code to allow users to easily extend the firmware, specifically using the new AI coding tools available.

As you have have already discovered, the memory constraints are quite tight on the AVR build, so I have to be very careful about which features get included in the final firmware

Claude's changes are inline with how I write the firmware but it's made a few changes which break compatibility. 

- Advancing the project version would prevent backwards compatability with MCL in general.
- Disabling WAV_DESIGNER is not desirable. 

Additionally, the least impactful way to get access to more AVR flash space, is to disable the Project Conversion/Upgrade features via the flag: 

> MCL_DISABLE_PROJECT_CONVERSION in mcl/MCL/MCLDefines.h:106

--

At the present moment I wont t be incorporating new functionality. I encourage you to continue to maintain your own fork of the repository to fit your specific needs.

Cheers

viktorakos95:
hey, quick update. i did both things you said, changed the wav designer disable to MCL_DISABLE_PROJECT_CONVERSION, and moved the settings to reserved[] instead of bumping project version, so that compatibility problem should be fixed now. its all on my fork, rebased on 5.02.

also i added one more thing: a Live Step Repeat/Roll (on the mixer page). You hold Left+Right arrows plus a pad and it rolls the sound at whatever subdivision you pick (select subdivision with Up/Down arrow), Also added an option on System menu if it should ignore Mute or not.

if you want, please feel free to implement the new one or the modified version of the previous one.

also, is it ok to mention these on discord or elektronauts, in case someone else also wants them from my fork?

jmamma:
Arrow key binding in mixer page conflicts with mute sets. But there is is probably another key combo that will work for beat repeat. maybe trig + scale ?

viktorakos95:
yeah it does interfere technically, but in practice it doesnt really affect the mute pages. actually you can use it together (combined with the added function of "roll ignores mute"): 

 1. hold left (goes into mute set 1 preview),
 2. while still holding left, also hold right (this arms the roll, still in the mute preview), 
 3. do the roll, 
 4. press yes (applies that mute preset). 
 
 so its kind of useful having it on the arrow keys, not just a conflict.

 if you still prefer scale key, i can switch to that, but scale already has a use there too for the secondary device mixer, so not fully free either.

 also just curious, are you actually leaning towards adding this, or just pointing out the conflict either way? dont want to change the key combo if its not going in anyway, since it works fine for me as is.

viktorakos95:
Hey,

I ended up also building an Elektron-style (Digitone II/Analog Rytm) Euclidean sequencer too for Machinedrum tracks. Two pulse generators with their own rotation, a boolean combine, track rotation, on/off toggle. Live preview until you commit it into real steps. Lives in the SEQ menu. Measured it in isolation and it's about 1.8KB flash / ~115 bytes RAM.

One heads up on fit: in my build, with Manual Step + Roll + Euclidean all together, I had to turn off project conversion and the file browser's MOVE too. If you're just looking at Euclidean on its own though, it'd probably fit with only project conversion off, no need to touch MOVE.

Happy to open a PR with just this if you're interested.

Also, in case you're not looking to implement it yourself, is it alright if I share what I've built on the Discord?

